PR #1, opened 2021-03-21
chore: add .env.example and a small startup instruction

smeijer:
I think this small change will make it easier to get started with this repo.


kentcdodds (approved):
Thanks!

PR #2, opened 2021-03-21
feat: don't purge tailwind in development mode

smeijer:
This change prevents tailwind (postcss) from purging css in development mode. Making things faster, and reduce the need for full server restarts.


kentcdodds (approved):
Ah! I didn't realize this was happening. Thanks!

PR #6, opened 2021-03-25
chore: rename .npmrc to .npmrc.example

smeijer:
I've set my Remix key in `~/.npmrc`, and because of that, I don't need this key once more in the local `.npmrc`.

It's not a big issue to have it twice, but... this env key in `.npmrc` also creates a dependency between the presence of that key, and the ability to run `git commit`. The git hooks trigger npm, which then reads `.npmrc`, which then fails if `REMIX_REGISTRY_TOKEN` is not set.

Long story short, without this change, people that wish to fix the frontend via the 'storybook route', will be unable to run `git commit`.


smeijer:
Actually, one small note before you hit merge :grimacing:

I think this will kill `npm i` in the GitHub actions? How can we fix that, without having that env key in the main `.npmrc`?

https://github.com/kentcdodds/remix-kentcdodds/blob/27a34aa04159b5a0641b6da5de8e7526de5841b5/.github/workflows/validate.yml#L26-L27

kentcdodds:
I'm thinking that the storybook idea won't work long-term (in fact, I'm pretty sure it doesn't work now), so I'm not worried about that (also, folks could use `--no-verify` if they really needed to).

The way it's currently set up was the simplest way to make everything work with deploys. I'm eventually planning on getting off of heroku and then we may have a different way to do all of this.

So for now, we'll leave this as-is :) Thanks though!

PR #11, opened 2021-06-10
feat: add dark-mode switch to storybook

smeijer:
Added `storybook-dark-mode`, so we can switch between light and dark mode in storybook.

It defaults to `dark` and applies the `dark` class to the `body` inside the iframe. I use the `storybook.css` file to control the frame's background color. These values should be adjusted to match the real theme when the style guide/color scheme comes in.

Preview:

storybook-dark-light


smeijer:
The 16k changes in `package-lock.json` are a bit annoying. Should we sync up npm versions? What version do you use?

kentcdodds (approved):
Cool!

PR #12, opened 2021-06-10
feat: serve static files via storybook

smeijer:
Added the `-s` flag to storybook, so that all static files from `/public` are available to it. This helps serving local images for example.

https://storybook.js.org/docs/riot/configure/images-and-assets#serving-static-files-via-storybook

Worth considering, I've only specified the `/public` folder, but we could also add a second folder like `/storybook/fixtures`. This folder could then be used for custom placeholder images that shouldn't be included on production servers.

### DIFF
--- a/storybook/pm2.config.js
+++ b/storybook/pm2.config.js
@@ -5,7 +5,7 @@ module.exports = {
     {
       name: 'Storybook',
       script:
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
👍

PR #13, opened 2021-06-18
feat: create home page

smeijer:
The PR seems to be quite large with 1547 lines, but as it's pretty much all visual work and everything is done via stories. So I think it's manageable. All the components are related to the home page.

## Dependencies

I've added:

- [`clsx`](http://npmjs.com/clsx) to have a way to dynamically construct classnames other than using string concatenation. 

- [`framer-motion`](http://npmjs.com/framer-motion) to add animations to some elements, so far only to the arrows in the problem/solution section.

- [`@reach/tabs`](http://npmjs.com/@reach/tabs) to have accessible tabs in the problem/solution section.

## Structure

**/components/icons**
I've grouped icon and logo components under this folder. Should we continue doing so? Or just move everything in the same `/components` root? While talking about icons. Do you prefer to inline svg's as I did here, or point an `<img` element to a remote `.svg` file instead?

Inline SVG's means fewer network requests, but a larger byte size for the first one. Remote SVG's means more requests, but better cachable. The cache benefit might be negligible.

**/components/sections**
In this folder, I've put the sections (building blocks) which together form pages. Sections don't have their own border margins. Instead, I put "spacer elements" between them when put on the page.

**/components/pages**:
A temporary folder where the "home page" currently lives. This content should be extracted and used within Remix, but this worked a bit faster during this first setup. Ideally, I wouldn't have this page here and have this `jsx` directly in a story instead. But as tailwind currently doesn't extract classNames from `*.stories.tsx` files, that didn't work, and I was running out of time.

**/public/placeholders**:
Placeholder images. We shouldn't be merging this PR before deleting those images from the repo, (and rebase so they were never here). What's your preferred way to deal with this kind of things? Just load images from `unsplash`? Or should we host them somewhere else?

The images don't play nice with the [netlify preview](https://deploy-preview-13--kcd-react.netlify.app/iframe.html?id=pages--home-page&args=&viewMode=story).

**/public/fonts**:
Should we host these on a CDN directly and point the css file to those? Or just depend on proper cache headers?


app/components/article-card.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function ArticleCard({
+  title,
+  date,
+  readTime,
+  imageUrl,
+  articleUrl,
+}: ArticleCardProps) {
+  const dateString = new Date(date).toLocaleDateString('en-US', {
```
smeijer, line 19: Should we extract date formatting to a util, and maybe make it smarter?

app/components/article-card.tsx
```diff
@@ -0,0 +1,43 @@
...
+    dateStyle: 'long',
+  })
```
smeijer, line 21: Should we extract date formatting to an util, and maybe make it smarter?

app/components/course-card.tsx
```diff
@@ -0,0 +1,49 @@
...
+}: CourseCardProps) {
+  return (
+    <div className="relative pt-12 w-full h-full">
+      <a
+        className="group relative block pb-10 pt-36 px-8 w-full h-full dark:bg-gray-800 rounded-lg md:pb-20 md:px-16"
+        href={courseUrl}
+      >
+        <div>
+          <H2>{title}</H2>
+        </div>
```
smeijer, line 27: Note to self: delete the `<div>`

app/components/footer.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+function GithubLink() {
+  return (
+    <a href="https://kcd.im/github">
+      <svg className="w-8 h-8" viewBox="0 0 24 24">
+        <path
+          fill="currentColor"
+          d="M12,2A10,10 0 0,0 2,12C2,16.42 4.87,20.17 8.84,21.5C9.34,21.58 9.5,21.27 9.5,21C9.5,20.77 9.5,20.14 9.5,19.31C6.73,19.91 6.14,17.97 6.14,17.97C5.68,16.81 5.03,16.5 5.03,16.5C4.12,15.88 5.1,15.9 5.1,15.9C6.1,15.97 6.63,16.93 6.63,16.93C7.5,18.45 8.97,18 9.54,17.76C9.63,17.11 9.89,16.67 10.17,16.42C7.95,16.17 5.62,15.31 5.62,11.5C5.62,10.39 6,9.5 6.65,8.79C6.55,8.54 6.2,7.5 6.75,6.15C6.75,6.15 7.59,5.88 9.5,7.17C10.29,6.95 11.15,6.84 12,6.84C12.85,6.84 13.71,6.95 14.5,7.17C16.41,5.88 17.25,6.15 17.25,6.15C17.8,7.5 17.45,8.54 17.35,8.79C18,9.5 18.38,10.39 18.38,11.5C18.38,15.32 16.04,16.16 13.81,16.41C14.17,16.72 14.5,17.33 14.5,18.26C14.5,19.6 14.5,20.68 14.5,21C14.5,21.27 14.66,21.59 15.17,21.5C19.14,20.16 22,16.42 22,12A10,10 0 0,0 12,2Z"
+        />
+      </svg>
```
smeijer, line 130: Note to self, for consistency, this should be extracted to an "icon component" and placed under `/components/icons`. The link can then be inlined in the footer.

app/components/footer.tsx
```diff
@@ -0,0 +1,202 @@
...
+        <path
+          fill="currentColor"
+          d="M22.46,6C21.69,6.35 20.86,6.58 20,6.69C20.88,6.16 21.56,5.32 21.88,4.31C21.05,4.81 20.13,5.16 19.16,5.36C18.37,4.5 17.26,4 16,4C13.65,4 11.73,5.92 11.73,8.29C11.73,8.63 11.77,8.96 11.84,9.27C8.28,9.09 5.11,7.38 3,4.79C2.63,5.42 2.42,6.16 2.42,6.94C2.42,8.43 3.17,9.75 4.33,10.5C3.62,10.5 2.96,10.3 2.38,10C2.38,10 2.38,10 2.38,10.03C2.38,12.11 3.86,13.85 5.82,14.24C5.46,14.34 5.08,14.39 4.69,14.39C4.42,14.39 4.15,14.36 3.89,14.31C4.43,16 6,17.26 7.89,17.29C6.43,18.45 4.58,19.13 2.56,19.13C2.22,19.13 1.88,19.11 1.54,19.07C3.44,20.29 5.7,21 8.12,21C16,21 20.33,14.46 20.33,8.79C20.33,8.6 20.33,8.42 20.32,8.23C21.16,7.63 21.88,6.87 22.46,6Z"
+        />
+      </svg>
+    </a>
+  )
+}
+
+// mt-20 md:mt-36
```
smeijer, line 161: cleanup comment

app/components/footer.tsx
```diff
@@ -0,0 +1,202 @@
...
+function Footer() {
```
smeijer, line 162: @kentcdodds, tabbing in the footer component feels a bit counterintuitive on `dekstop` resolutions. Elements are rearranged when resizing the browser, but <kbd>tab</kbd> follows the DOM, and not the visual order. So on desktop, focus jumps from the social icons on the left, to the newsletter signup on the right, to the nav links in the center.

On smaller resolution, the focus follows the right path. I'm not sure what's the right thing to do here. I see two solutions:

1. Add javascript width detection, and enforce tab order that way (not a fan)
2. Duplicate the markup for the newsletter, hide the version out of range, and show the version in range. That means double HTML, but the focus works as expected. (not really awesome either)

Do you have other ideas/suggestions?

app/components/grid.tsx
```diff
@@ -0,0 +1,50 @@
...
+import clsx from 'clsx'
+import type {ReactNode} from 'react'
+
+interface GridProps {
+  children: ReactNode
+  overflow?: boolean
+  className?: string
+}
+
+function Grid({children, className}: GridProps) {
```
smeijer, line 11: The base layout component. So far, everything besides the navbar is wrapped in it. The column counts and gap sizes are the same as in the design files. Gotta :heart: CSS Grids!

app/components/grid.tsx
```diff
@@ -0,0 +1,50 @@
...
+        className,
+      )}
+    >
+      {children}
+    </div>
+  )
+}
+
+/**
+ * Use for development only! It renders the grid columns and gaps as page overlay
```
smeijer, line 25: Really useful while creating new components/sections. Just add this one to your markup, and you'll have a nice visual column grid. 

```jsx
<>
  <GridLines />
  <Grid>
    <div ...
  </Grid>
<>
```

app/components/navbar.tsx
```diff
@@ -0,0 +1,94 @@
...
+interface NavLinkProps {
+  href: string
+  active?: boolean
+  children?: ReactNode
+}
+
+function NavLink({href, children}: NavLinkProps) {
+  return (
+    <li>
+      <a
```
smeijer, line 16: I think we should replace this `<a>` with `Link` from `react-router-dom`, but storybook stopped working then. We should add a decorator to `preview.js` to wrap all stories in `react-router`.

app/components/navbar.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function DarkModeToggle() {
+  const [mode, setMode] = React.useState<'light' | 'dark'>('dark')
+
+  React.useEffect(() => {
+    const classList = document.body.classList
+    classList.remove('light', 'dark')
+    classList.add(mode)
+  }, [mode])
```
smeijer, line 33: The effect is just a quick hack, should be replaced someday.

app/components/navbar.tsx
```diff
@@ -0,0 +1,94 @@
...
+        </NavLink>
+        <NavLink href="/discord">Discord</NavLink>
+        <NavLink href="/podcast">Podcast</NavLink>
+        <NavLink href="/workshops">Workshops</NavLink>
+        <NavLink href="/about">About</NavLink>
+      </ul>
+
+      <div className="flex items-center justify-center">
+        <div className="block lg:hidden">
+          <MenuButton />
```
smeijer, line 76: This thing should toggle the mobile menu. Hasn't been implemented yet.

app/components/pages/home.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {Footer} from '../footer'
+
+// NOTE: The `pages` folder is only temporary. I've placed this here, because
+//   tailwind doesn't extract css from .storybook.tsx files. Which is fine, but I
+//   needed a quick fix.
+
+// TODO: I'm not sure if we should add the "spacer elements", or add the spacing
+//  inside the sections. There is a slight variation between element paddings,
+//  which might make it tricky to get it right when embedding inside sections,
+//  when sections turn out to be reused. Let's decide later what to do with this.
```
smeijer, line 21: Made a comment about this file in the PR body as well.

app/components/sections/blog-section.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <div className="flex flex-col col-span-full mb-10 space-y-10 lg:flex-row lg:items-center lg:justify-between lg:space-y-0">
+        <div className="space-y-2 lg:space-y-0">
+          <H2>Most popular from the blog.</H2>
+          <H2 variant="secondary">Probably the most helpful aswell.</H2>
+        </div>
+
+        <ArrowButton direction="right">See the full blog</ArrowButton>
+      </div>
+
+      {/*This should be a horizontal scroller */}
```
smeijer, line 24: note to self, no it shouldn't. I've talked to the designer about it, and the code has already been adjusted. Forgot to remove the comment :grimacing:. On desktop, it shows 3 columns. On tablet 2, and on mobile 2 rows.

app/components/sections/hero-section.tsx
```diff
@@ -0,0 +1,43 @@
...
+// Note that the image overlaps the right edge of the grid by `8vw`. This `8vw`
+// needs to stay in sync with the `10vw` margins of the grid component.
+function HeroSection() {
+  return (
+    <Grid className="h-full">
+      <div className="relative col-span-full px-4 lg:col-span-6 lg:col-start-7 lg:px-0 lg:h-full">
+        <div className="lg:right-[-8vw] bottom-0 left-0 right-0 top-0 flex items-center justify-center lg:absolute">
+          <img
+            alt=""
+            className="max-w-[80vh] w-full h-auto object-contain"
```
smeijer, line 16: This `max-w-[8-vh]` should probably be dropped. Need to spend a bit more time on this to get it right.

app/components/sections/problem-solution-section.tsx
```diff
@@ -0,0 +1,96 @@
...
+            <hr className="col-span-full mb-16 mt-20 border-gray-200 dark:border-gray-600" />
+
+            <div className="col-span-full order-2 mt-16 lg:col-span-5 lg:col-start-7 lg:mt-0">
+              <img className="-ml-10 w-48" src="/placeholders/skis.png" />
+            </div>
+
+            <div className="col-span-full col-start-1 order-1 lg:col-span-5 lg:order-3">
+              <ul className="flex flex-row text-white text-xl space-x-8 lg:flex-col lg:text-7xl lg:space-x-0 lg:space-y-7">
+                <li>
+                  <a
```
smeijer, line 42: I'm unsure what this menu is about, or what it does. It has an arrow, and it has 1 bright and 2 dimmed links. What does it do?

app/components/title.tsx
```diff
@@ -0,0 +1,78 @@
...
+  variant?: 'primary' | 'secondary'
+  children: ReactNode
+}
+
+const titleColors = {
+  primary: 'text-black dark:text-white',
+  secondary: 'text-gray-400 dark:text-blueGray-500',
+}
+
+function H1({variant = 'primary', children}: TitleProps) {
```
smeijer, line 15: The tag names are extracted from the designs. I'm not sure they reflect the proper semantics. Almost everything on the home page is a heading. What's wisdom here?

smeijer:
Also, hover, focus and active states are mostly absent. I think we have to add all those when most pages are done. Same for aria labels etc, I think it makes sense to add those while connecting the components to the backend and while making things interactive.

smeijer:
Regarding the fonts, I didn't see it on Google Fonts, so I converted the otf to woff myself. Do we need to include a license file for it in this repo?

app/components/footer.tsx
```diff
@@ -0,0 +1,202 @@
...
+  return (
+    <footer className="pb-16 pt-48 border-t border-gray-200 dark:border-gray-600">
+      <Grid className="grid-rows-max-content">
+        {/*about section*/}
```
kentcdodds, line 166: These comments can probably be removed now.

app/components/grid.tsx
```diff
@@ -0,0 +1,50 @@
...
+  children: ReactNode
+  overflow?: boolean
+  className?: string
+}
+
+function Grid({children, className}: GridProps) {
+  return (
+    <div
+      className={clsx(
+        'gap-[18px] mx-[10vw] md:gap-[17px] lg:gap-[23px] grid grid-cols-4 md:grid-cols-8 lg:grid-cols-12',
```
kentcdodds, line 15: I can't remember what this syntax is called, but if we can avoid it let's do in effort to keep sizes consistent. If it's a matter of the design being off from one of the pre-set sizes by a pixel or two then let's just go with the pre-set size rather than using this syntax. I'm concerned that overusing this reduces consistency.

app/components/grid.tsx
kentcdodds: Handy!

app/components/navbar.tsx
kentcdodds: Cool, good to know. I'll take care of this so you can keep pushing forward on what you're working on.

app/components/navbar.tsx
kentcdodds: I have done a bit of work on this but I'll want to change it I think. Here it is now: https://github.com/kentcdodds/remix-kentcdodds/blob/de8f7ec532d414cfbc2ad1798be16cd28032d734/app/theme-provider.tsx

I think this is something I can handle.

tailwind.config.js
```diff
@@ -10,6 +10,73 @@ module.exports = {
...
+      },
+      blueGray: {
+        500: '#A9ADC1',
+      },
+      // I'm not sure if I'm a fan of `team-x`, but team-blue is close to `blue-500`. Adding
+      // this to blue-400 or blue-600, leads to confusion
+      team: {
+        yellow: '#FFD644',
+        blue: '#36A3FF',
+        red: '#FF4545',
```
kentcdodds, line 38: I'm a fan 👍

app/components/pages/home.tsx
kentcdodds: I think I'm in favor of `<Spacer />` or a `<Stack />` component like in [Seek's Braid Design System](https://seek-oss.github.io/braid-design-system/components/Stack).

app/components/sections/about-section.tsx
```diff
@@ -0,0 +1,38 @@
...
+          src="/placeholders/about-picture.png"
+        />
+      </div>
+
+      <div className="flex flex-col col-span-full justify-center space-y-12 lg:col-span-5 lg:col-start-8 lg:mt-0">
+        <div>
+          <img className="w-auto h-48" src="/placeholders/snowboard.png" />
+        </div>
+
+        <H2>Big extreme sports enthousiast. </H2>
```
kentcdodds, line 22:
```suggestion
        <H2>Big extreme sports enthusiast.</H2>
```

app/components/sections/blog-section.tsx
```diff
@@ -0,0 +1,39 @@
...
+  articles: ArticleCardProps[]
+}
+
+function BlogSection({articles}: BlogSectionProps) {
+  return (
+    <Grid>
+      <div className="flex flex-col col-span-full mb-10 space-y-10 lg:flex-row lg:items-center lg:justify-between lg:space-y-0">
+        <div className="space-y-2 lg:space-y-0">
+          <H2>Most popular from the blog.</H2>
+          <H2 variant="secondary">Probably the most helpful aswell.</H2>
```
kentcdodds, line 18:
```suggestion
          <H2 variant="secondary">Probably the most helpful as well.</H2>
```

app/components/sections/problem-solution-section.tsx
kentcdodds: I *think* that these should be `<button>`s and clicking them will update the info on the right. Could/should probably use [`@reach/tabs`](https://reach.tech/tabs) for this.

app/components/title.tsx
kentcdodds: If it's just styled like a heading but isn't semantically a heading that's fine. Let's just make sure the semantics are correct. Maybe we could support an `as` prop (I normally hate those kinds of props, but I think it makes sense here). So folks could do: `<H1 as="div" ... />`

kentcdodds (commented):
Love this so much! Thanks for what you've got so far. Left some (hopefully) helpful comments. Great direction here 👍

kentcdodds:
> I've grouped icon and logo components under this folder. Should we continue doing so? 

Yes, that's great 👍 In general I like keeping folder structure pretty flat and only splitting files when it's really necessary. But when there's an obvious grouping then that's fine.

kentcdodds:
> Do you prefer to inline svg's as I did here

Definitely.

For placeholders, feel free to use unsplash if you're worried about license issues. I'll get some real images soon.

For font, let's keep it as-is and I'll optimize it later 👍

app/components/footer.tsx
smeijer: That should work. That's what I meant with "2". Let's do that.

app/components/sections/problem-solution-section.tsx
smeijer: That idea had crossed my mind. Let's make it like that.

app/components/course-card.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <div className="mt-4 max-w-sm">
+          <H2 variant="secondary">{description}</H2>
```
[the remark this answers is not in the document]
smeijer: Agreed. Let's implement that `as` prop, and get this right.

app/components/grid.tsx
smeijer: I agree. I'll change it.

Regarding the grid. I think we should add a max-width. Things get weird when making the browser 3000px wide for example.

app/components/sections/problem-solution-section.tsx
kentcdodds: Would be cool if when you changed the selection there was a tiny animation of the arrow to move slightly from the left to the right

kentcdodds:
> Do we need to include a license file for it in this repo?

Go ahead and put a LICENSE file in the directory next to the font files just to be safe :)

kentcdodds:
Just force-pushed a rebase on `main` so the github action build works (hopefully) 😅

kentcdodds:
Ok, I fixed it for realz this time :)

kentcdodds:
https://github.com/kentcdodds/remix-kentcdodds/commit/5df61e6d4e737bbc1720624b1017cfb7e5b52dae :)

app/components/article-card.tsx
smeijer: Added to a project board so it's not forgotten. Let's make this part of another PR where it's more relevant.

https://github.com/kentcdodds/remix-kentcdodds/projects/1#card-63944862

app/components/navbar.tsx
smeijer: Added a note to the project. Let's do this when we're connecting things to Remix, and know where the links should point to.

https://github.com/kentcdodds/remix-kentcdodds/projects/1#card-63944993

app/components/sections/problem-solution-section.tsx
smeijer: Done. Might need some fine-tuning in speed and distance, but that's trivial.

https://github.com/kentcdodds/remix-kentcdodds/blob/f31b3fefa09fc56cac7768b2574999e2b5b60f20/app/components/sections/problem-solution-section.tsx#L33-L35

app/components/navbar.tsx
smeijer: Created a card https://github.com/kentcdodds/remix-kentcdodds/projects/1#card-63945001

smeijer:
I've resolved most of the todos, and others have been moved to [the project board](https://github.com/kentcdodds/remix-kentcdodds/projects/1) that I've created. I think it's good to add notes there so we don't forget things and can move some "todos" to it, so we can move forward to maintain speed and stick to planning.

We can then prioritize todos there, and either fix them directly, dispose them, or convert them to issues when they're moved to the "long term".

I've also:

- decided to not replace the `<a>` with `<Link>` at this moment, as I believe that it makes more sense to do that once we connect the components to Remix / create the real pages.

- created the `Spacer` & `Stack` components and decided to use `Spacer` because that felt like a better match due to differences in space between certain sections. Stack groups sections (in code) based on the space between them, which looked clean, but felt confusing in this case. I've still committed `Stack`, maybe we see a purpose for it on other pages, maybe not. Delete is one click away.

- removed the tailwind bracket syntax, merged some props like the grid-gap between `sm` and `md` breakpoints, as they only had a 1px difference. For other things, I've added some utils via `tailwind.config.js` (such as the page margin, hero offset, and arrow rotation). There is now not a single instance of bracket notation in this PR, which should improve maintainability.

- replaced most placeholder images with images from unsplash, or images from https://epicreact.dev. There are a few left, but I think those are fine unless we're able to temporarily host them somewhere else. I didn't replace those three with other random assets, because they are an essential part of the design.

- I've refactored the `H1-6` components a bit, so they support the `as` prop. We can turn all those different components (title & paragraph) into a single generic `Typography` component but I didn't want to spend time on it at this moment. I've created [note for it](https://github.com/kentcdodds/remix-kentcdodds/projects/1#card-63944823), and would love to hear your thoughts about that. 

  My idea was that we could create something like:

  ```jsx
  <>
    <Typography size="xl" color="secondary" tagName="h1" />
    <Typography size="md" color="primary" tagName="p" />
  </>
  ```

**Note: ** When merging, use `squash merge` so that the removed images won't be stored forever.

smeijer:
Force pushed because I rebased on main to fix conflicts in `package-lock.json`

kentcdodds:
This is great! Good work on this so far 👍

I'm going to merge this now. We can clean up stuff later 👍

image

PR #15, opened 2021-06-28
fix: force vertical alignment of discord speech bubbles

smeijer:
Fixes the speech bubble as shown in: https://github.com/kentcdodds/remix-kentcdodds/pull/13#issuecomment-869834796

image


kentcdodds (approved):
👍

PR #16, opened 2021-06-28
fix: polish home page

smeijer:
- changed hero to match screen height, without the navbar+hero wrapper.
- fixed the `course-card` to have a background on light mode
- fixed the `@reach/tabs` to have correct styles when `@reach/tabs/style.css` is imported, by changing the app.css and tailwind.css load order
- fixed page background color for light/dark modes

### DIFF
--- a/app/components/sections/hero-section.tsx
+++ b/app/components/sections/hero-section.tsx
@@ -7,10 +7,12 @@ import {Button} from '../button'
 // Note that the image overlaps the right edge of the grid by `8vw`. This `8vw`
 // needs to stay in sync with the `10vw` margins of the grid component.
 function HeroSection() {
+  // The grid has a computed height on the large breakpoint, to make the Hero
+  // span the height of the screens, minus 10rem for the navbar (h-40).
   return (
-    <Grid className="h-full">
+    <Grid className="lg:h-[calc(100vh-10rem)] lg:pb-12">
       <div className="relative col-span-full px-4 lg:col-span-6 lg:col-start-7 lg:px-0 lg:h-full">
-        <div className="bottom-0 left-0 right-0 top-0 flex items-center justify-center lg:absolute lg:-right-8vw">
+        <div className="bottom-0 left-0 right-0 top-0 flex items-center justify-center pointer-events-none lg:absolute lg:-right-8vw">
           <img
             alt="A mascot standing on a snowboard surrounded by green leaves, a battery, two skies, a one-wheel, a solar panel and a recycle logo."
             className="w-full h-auto max-h-screen object-contain"
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -28,6 +28,7 @@ import {getEnv} from './utils/env.server'
 import {Navbar} from './components/navbar'
 import {Spacer} from './components/spacer'
 import {Footer} from './components/footer'
+import clsx from 'clsx'
 
 export const meta: MetaFunction = ({data}: {data: LoaderData}) => {
   const theme = getThemeFromMedia(data.theme)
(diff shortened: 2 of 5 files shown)


app/components/sections/hero-section.tsx
```diff
@@ -7,10 +7,12 @@ import {Button} from '../button'
 // Note that the image overlaps the right edge of the grid by `8vw`. This `8vw`
 // needs to stay in sync with the `10vw` margins of the grid component.
 function HeroSection() {
+  // The grid has a computed height on the large breakpoint, to make the Hero
+  // span the height of the screens, minus 10rem for the navbar (h-40).
   return (
-    <Grid className="h-full">
+    <Grid className="lg:h-[calc(100vh-10rem)] lg:pb-12">
```
smeijer, line 13: This is indeed a bracket thing while I just removed all of those in another commit. But I think it's the best way to do this. We need to subtract the navbar height from the screen height, to get the hero height.

app/components/sections/hero-section.tsx
```diff
@@ -7,10 +7,12 @@ import {Button} from '../button'
...
       <div className="relative col-span-full px-4 lg:col-span-6 lg:col-start-7 lg:px-0 lg:h-full">
-        <div className="bottom-0 left-0 right-0 top-0 flex items-center justify-center lg:absolute lg:-right-8vw">
+        <div className="bottom-0 left-0 right-0 top-0 flex items-center justify-center pointer-events-none lg:absolute lg:-right-8vw">
```
smeijer, line 15: We need to disable pointer-events on this image. Not sure yet why or how, but it was blocking click events on the dark/light toggle. Might have been something locally, I can look into it later.

app/components/sections/hero-section.tsx
kentcdodds: Yup, I have no problem with brackets for situations like this 👍

app/components/sections/hero-section.tsx
kentcdodds: We need to figure out how to avoid it covering more than the space it's allowed:

image

app/root.tsx
```diff
@@ -90,10 +91,15 @@ function App() {
         <Links />
       </head>
       <body
-        // TODO: switch this to clsx when Stephan's PR that adds it is merged
-        className={[showPendingState ? 'opacity-50' : null, theme]
-          .filter(Boolean)
-          .join(' ')}
+        className={clsx(
+          'transition',
```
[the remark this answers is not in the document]
kentcdodds: Theme stuff is kinda wonky because we need to server render the right theme, but we don't know what the right theme is on the server. I'll look into this more 👍

app/root.tsx
```diff
@@ -52,8 +53,8 @@ export const meta: MetaFunction = ({data}: {data: LoaderData}) => {
 export const links: LinksFunction = () => {
   return [
     {rel: 'icon', href: '/favicon.ico'},
-    {rel: 'stylesheet', href: tailwind},
     {rel: 'stylesheet', href: styles},
+    {rel: 'stylesheet', href: tailwind},
```
kentcdodds, line 57: This seems wrong. We should make our styles be able to override what's in tailwind right?

kentcdodds (approved):
Looks good to me. I'll go ahead and merge it as-is, but I'm still confused by the tailwind ordering there.

app/root.tsx
smeijer: Yeah, I'm indeed also not 100% sure. An alternative would be to add things like `@reach` to a `vendors.css` file, that we load very first. We need to be able to override `@reach`, which is currently loaded via `app`.

That being said, this is unlikely to give trouble, because if we would need to overrule tailwind, we could write more specific selectors.

kentcdodds: Ah, yeah, we should make a `vendors.css` file 👍 I'll handle that. Thanks!

kentcdodds: Alrighty, I've fixed this up well :)

PR #19, opened 2021-06-30
feat: add blog detail page

smeijer:
I've added some typography (prose) styles to `tailwind.config.js` so we control blog styles through prose. This felt better than providing custom components to the mdx parser. Prose for blogs, utility classes for everything else.

I've also fixed a small typo in the `KCDMdxComponent.components` prop, and I've removed the row-gap from the grid component. The column-gap works great, but the row-gap was only standing in the way.

I'm not done yet. I need to add a few more element types, and I'm chatting with the designer about some details. So

<p align="center">
to be continued
</p>

## TODO

- [ ] fix candy-dispenser `/blog/usememo-and-usecallback`
- [ ] what to do with the transparent `png` at `/blog/unit-vs-integration-vs-e2e-tests` ?

### DIFF
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -155,7 +155,7 @@ function Footer() {
           <ContactSection />
         </div>
 
-        <div className="col-span-2 mt-20 md:col-start-7 md:row-start-1 md:mt-0 lg:col-start-5 lg:row-start-2 lg:mt-10">
+        <div className="col-span-2 mt-20 md:col-start-7 md:row-start-1 md:mt-0 lg:col-start-5 lg:row-start-2 lg:mt-16">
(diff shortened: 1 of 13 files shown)


smeijer:
Regarding the candy dispenser, do a lot of posts have embedded jsx/html? We could adjust the prose classes to either do something with embedded html, or attempt to not target it entirely. But I feel like content like that should be positioned in an iframe, with its own styling. 

Otherwise, we'll always have the chance of missing some (edge) cases.

smeijer:
Despite the open issues in this PR, I'm okay with getting it merged so that Gil look at some articles on staging. Ready when you are.

kentcdodds (approved):
Solid progress here 👍

PR #22, opened 2021-07-01
small improvements for blog index page

smeijer:
Hide the `Search blog by topics` caption when there are no topic buttons to click on, and move around some spacing to make everything right.


kentcdodds (approved):
👍 there won't be no topics in the future most likely, but I don't mind coding defensively here

PR #24, opened 2021-07-01
feat: add login page

smeijer:
Polished the login page :sunglasses:


kentcdodds:
Looks super! I'll look at the code soon, but I had a few thoughts on the visuals.

Can you figure out why autofill makes it look so bad (both on light and dark mode):

image

I'm not 100% certain I prefer this over the gray (teamless) Alex profile. What do you think?

image

image

Rather thank making the entire form disappear, I *think* I'd rather render this message in the same place we put the error message:

image

I especially like including the ✨ because the magic link email includes it so it helps create cohesion. What do you think?

kentcdodds:
One other note about the teamless Alex profile photo. When a user is logged in, if they don't have a photo then we fallback to Alex in their team color, so I think that creates cohesion.

smeijer:
True, but it can also be confusing. It might give the user the impression they're logged in when they're not. Let's roll back Alex, and see if we get messages from confused users?

Regarding the autofill thing, I'll try fix it, but it will be hacky. Browsers don't really support customizing the autofill style. Which sucks.

smeijer:
Regarding Alex again, desktop resolution labels that button as `Sign up` (I need to add that). That makes switching to Alex on mobile resolution a bit confusing as well. But the user icon doesn't really feel like a "sign up" either :thinking: .

image

kentcdodds:
I think that design was made before we had avatars

smeijer:
- overruled the autofill colors
- restored Alex on both mobile and desktop resolutions
- rendered the `success message` inline

app/components/icons/user-icon.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react'
+
+function UserIcon() {
```
kentcdodds, line 3: We can delete this I think.

app/components/form-elements.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react'
+
+function Label(props: JSX.IntrinsicElements['label']) {
+  return (
+    // eslint-disable-next-line jsx-a11y/label-has-associated-control
```
kentcdodds, line 5: Maybe we should disable this rule. It's a pain and rarely actually helps in real-world situations like this.

styles/app.css
```diff
@@ -1,3 +1,30 @@
 :focus:not(:focus-visible) {
   outline: none;
 }
+
```
kentcdodds, line 4: lol, what a mess we have to deal with here 😆

Thanks!

styles/app.css
```diff
@@ -1,3 +1,30 @@
...
+input:-webkit-autofill,
+input:-webkit-autofill:hover,
+input:-webkit-autofill:focus,
+input:-webkit-autofill:active {
+  -webkit-text-fill-color: black !important;
+  -webkit-box-shadow: 0 0 0 999px #e6e9ee inset !important; /* gray-200 */
```
kentcdodds, line 10: Maybe we should create a css variable so this and `gray-200` can use the same thing (also allows us to change it for light/dark mode as needed).

styles/app.css
```diff
@@ -1,3 +1,30 @@
...
+  background-clip: content-box !important;
+}
+
+.dark input:-webkit-autofill,
+.dark input:-webkit-autofill:hover,
+.dark input:-webkit-autofill:focus,
+.dark input:-webkit-autofill:active {
+  -webkit-text-fill-color: white !important;
+  -webkit-box-shadow: 0 0 0 999px #2e3039 inset !important; /* gray-800 */
```
kentcdodds, line 19: Same here?

tailwind.config.js
```diff
@@ -44,6 +44,10 @@ module.exports = {
         500: '#4B96FF',
         100: '#E8F2FF',
       },
+      red: {
+        // TODO: decide if this is a good color, I just took team-red
+        500: '#FF4545',
```
kentcdodds, line 49: Seems legit to me. I'd like to hear what Gil has to say though 👍

kentcdodds (commented):
Looking great!

kentcdodds:
Sorry about the conflict. I resolved it for you.

app/components/form-elements.tsx
smeijer: Done

styles/app.css
smeijer: Somewhat done with ae287d39f8f2940264167f65bf79cadf7cf26ca6. I think we can build upon that to improve things futher. Think of slimming down the gray-pallete, and assign colors like. I need to research a bit, but I think we can reduce our current 8 gray colors, to 4 (per theme) by doing so.

```jsx
<div className="bg-gray-dark" />
```

instead of:

```jsx
<div className="bg-gray-200 dark:bg-gray-800" />
```

kentcdodds (approved):
Super duper 👍

PR #26, opened 2021-07-02
feat: implement signup page

smeijer:
This one is nice!

- added some colors & content to the signup page
- added `team-current` as tailwind color, which allows us to do things like `text-team-current`, `bg-team-current` or `focus:ring-2 ring-team-current` :sunglasses:
- those focus styles, are live updated on the signup page. Selecting a team, changes the color of the name input :fire:


cypress/e2e/onboarding.ts
```diff
@@ -30,7 +30,8 @@ describe('onboarding', () => {
     cy.findByRole('main').within(() => {
       cy.findByRole('textbox', {name: /name/i}).type(firstName)
       cy.findByRole('group', {name: /team/i}).within(() => {
-        cy.findByRole('radio', {name: /blue/i}).click()
+        // checkbox is covered with a <label>
+        cy.findByRole('radio', {name: /blue/i}).click({force: true})
```
smeijer, line 34: The radio is wrapped in a label, but that's a good thing. Right?

```
  1) onboarding
       should allow a user to register a new account:
     CypressError: Timed out retrying after 4050ms: `cy.click()` failed because this element:

`<input type="radio" class="sr-only" name="team" value="BLUE">`

is being covered by another element:

`<label class="block pb-12 pt-20 px-12 text-center">...</label>`

Fix this problem, or use {force: true} to disable error checking.
```

cypress/e2e/onboarding.ts
```diff
@@ -30,15 +30,18 @@ describe('onboarding', () => {
...
+      // cy.findByRole('button', {name: /logout/i}).click()
     })
 
-    cy.findByRole('navigation').within(() => {
-      cy.findByRole('link', {name: /login/i})
-    })
+    // TODO: find a new home for the login-with-qr-code thing?
+    // cy.findByRole('navigation').within(() => {
+    //   cy.findByRole('link', {name: /login/i})
+    // })
```
smeijer, line 45: Ok. These are not so awesome. But the current designs haven't taken them into account, and I lack the creativity to come up with something nice. I'll ask Gil.

smeijer: Oh wait, I'm mixing stuff up, aren't I. This is `login link`, shown on the signup page, not on the `/me` page. So that's not the QR. What is it?

smeijer: aah, I'm mixing something up. This should be the main nav link, right? Let's fix that one.

cypress/e2e/onboarding.ts
```diff
@@ -30,11 +30,13 @@ describe('onboarding', () => {
...
-        cy.findByRole('radio', {name: /blue/i}).click()
+        // checkbox is covered with a <label>
+        cy.findByRole('radio', {name: /blue/i}).click({force: true})
       })
-      cy.findByRole('button', {name: /join KCD/i}).click()
+      cy.findByRole('button', {name: /create account/i}).click()
 
-      cy.findByRole('button', {name: /logout/i}).click()
+      // TODO: find a new home for the logout button?
+      // cy.findByRole('button', {name: /logout/i}).click()
```
smeijer, line 39: Ok. This is not so awesome. But the current designs haven't taken them into account, and I lack the creativity to come up with something nice. I'll ask Gil.

smeijer: Ok. Gonna try something while I wait for the designer. Got some inspiration.

smeijer: Done.

image

app/routes/signup.tsx
```diff
@@ -135,61 +144,209 @@ export const loader: LoaderFunction = async ({request}) => {
...
+                  You can’t change this later.
+                </H2>
+              </div>
+              <Link
+                to="/logout"
+                className="flex col-span-full items-center px-11 py-6 text-black dark:text-white border-2 border-gray-600 rounded-full space-x-4 lg:col-span-8 lg:col-start-3"
+              >
+                <LogoutIcon />
+                <H6 as="span">logout</H6>
+              </Link>
```
kentcdodds, line 252: They're not actually logged in until they've created an account so I'm not sure it makes sense to include the logout button here 🤔

smeijer: Happy to remove it. Just say the word. 😅

kentcdodds: Yeah, let's do. We can use it on the profile page though

app/routes/signup.tsx
```diff
@@ -135,61 +142,198 @@ export const loader: LoaderFunction = async ({request}) => {
   })
 }
 
+const TEAM_MAP = {
```
kentcdodds, line 145: Why don't we do something similar to this: https://github.com/kentcdodds/remix-kentcdodds/blob/781a1f33612f43514eb2de6f63f3ea61c6740fd0/app/utils/misc.tsx#L26-L30

```suggestion
const TEAM_MAP: Record<Team, string> = {
```

app/routes/signup.tsx
```diff
@@ -135,61 +142,198 @@ export const loader: LoaderFunction = async ({request}) => {
...
+  },
+  YELLOW: {
+    image: images.alexYellow,
+    label: 'Yellow Team',
+    focusClassName: 'ring-team-yellow',
+  },
+}
+
+interface TeamOptionProps {
+  team: keyof typeof TEAM_MAP
```
kentcdodds, line 164: With the change above, if this isn't happy, I think we could just do this:

```suggestion
  team: Team
```

app/routes/signup.tsx
```diff
@@ -135,61 +142,198 @@ export const loader: LoaderFunction = async ({request}) => {
...
-            >
-              {data.errors.firstName}
+    <div className="mt-12">
+      <style>{`
+          :root {
+            --color-team-current: var(--color-team-${(
+              formValues.team || 'UNKNOWN'
+            ).toLowerCase()}); 
+          }
+        `}</style>
```
kentcdodds, line 223: Oh, clever idea! Let's add a comment explaining what this is doing and why we can't use the `--color-team-current` that's defined in the `root.tsx` in case I come back to this in a few months and I forget 😅

app/routes/signup.tsx
```diff
@@ -135,61 +142,198 @@ export const loader: LoaderFunction = async ({request}) => {
...
+        },
+      )}
+    >
+      {selected ? (
+        <span className="absolute left-9 top-9 text-team-current">
+          <CheckIcon />
+        </span>
+      ) : null}
+
+      <label className="block pb-12 pt-20 px-12 text-center">
```
kentcdodds, line 189: Let's add a `cursor-pointer` to this. Otherwise people may not realize they can select it.

app/routes/signup.tsx
```diff
@@ -135,61 +142,198 @@ export const loader: LoaderFunction = async ({request}) => {
...
 export default function NewAccount() {
   const data = useRouteData<LoaderData>()
+  const [formValues, setFormValues] = React.useState({
+    firstName: '',
+    team: '',
+  })
+
+  const formIsValid =
+    formValues.firstName.trim().length > 0 &&
+    Object.keys(TEAM_MAP).includes(formValues.team)
```
kentcdodds, line 213: If we [move this](https://github.com/kentcdodds/remix-kentcdodds/blob/781a1f33612f43514eb2de6f63f3ea61c6740fd0/app/utils/prisma.server.ts#L196) to `misc.tsx` (so it's not in a `.server.ts` file), then we could do this:

```suggestion
    teams.includes(formValues.team)
```

cypress/e2e/onboarding.ts
kentcdodds: That's fine 👍 It's accessible and that's all that matters.

app/routes/signup.tsx
kentcdodds: On second thought. Do you think that we could use a `useEffect` instead that does this: 

```tsx
document.documentElement.style.setProperty('--color-team-unknown', `var(--color-team-${(formValues.team || 'UNKNOWN').toLowerCase()})`)
```

That way the circle in the top right updates too:

image

And actually it just occurred to me that we know enough by this time to get the user's avatar as well, so I'll work on that bit.

kentcdodds (commented):
Super work here! I'm loving this page 👍

Just a few comments.

app/routes/signup.tsx
kentcdodds: I'm moving that now myself 👍

app/routes/signup.tsx
kentcdodds: Starting to think I wanna put this in context so the avatar can be updated too...

kentcdodds: Go ahead and leave this as-is. I'll come back around to it.

app/routes/signup.tsx
kentcdodds: Actually, this:

```suggestion
const TEAM_MAP: Record<Team, {
    image: string,
    label: string,
    focusClassName: string,
  }> = {
```

It's mostly just so we make sure we don't have a typo on the team name today or in the future.

smeijer: Yeah, agree. This should be typed.

kentcdodds (approved):
This is great work 👏

PR #27, opened 2021-07-02
various optimizations to code blocks

smeijer:
Improved the highlighted code blocks:

- code blocks now span full screen width on narrow screens (`< screen:lg`)
- highlighted lines are now extended to the full frame width, also when horizontally overflown
- line numbers are now sticky, so that they'll remain visible when scrolling horizontally


kentcdodds (approved):
Sweeeet! Love it :)

PR #28, opened 2021-07-02
feat: implement profile page

smeijer:
I've added some colors and content to the profile page, but I'm waiting for response from Gil for the QR code implementation.

### DIFF
--- a/app/components/icons/logout-icon.tsx
+++ b/app/components/icons/logout-icon.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react'
+
+function LogoutIcon() {
+  return (
+    <svg width="24" height="24" fill="none" viewBox="0 0 24 24">
+      <path
+        stroke="currentColor"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        strokeWidth="1.5"
+        d="M15.75 8.75L19.25 12L15.75 15.25"
+      />
+      <path
+        stroke="currentColor"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        strokeWidth="1.5"
+        d="M19 12H10.75"
+      />
+      <path
+        stroke="currentColor"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        strokeWidth="1.5"
+        d="M15.25 4.75H6.75C5.64543 4.75 4.75 5.64543 4.75 6.75V17.25C4.75 18.3546 5.64543 19.25 6.75 19.25H15.25"
+      />
+    </svg>
+  )
+}
+
+export {LogoutIcon}
--- a/app/routes/me.tsx
+++ b/app/routes/me.tsx
@@ -117,6 +117,8 @@ function YouScreen() {
               required
               defaultValue={user.email}
               className="mb-8"
+              readOnly
+              disabled
             />
 
             <div className="flex flex-wrap items-baseline justify-between mb-4">
@@ -133,7 +135,10 @@ function YouScreen() {
                     connected
                   </a>
                 ) : (
-                  <a className="text-black dark:text-white" href={authorizeURL}>
+                  <a
+                    className="focus-ring text-black dark:text-white rounded-lg"
+                    href={authorizeURL}
+                  >
                     Connect to Discord
                   </a>
                 )}
@@ -145,9 +150,10 @@ function YouScreen() {
               name="discord"
               value={userInfo.discord?.username ?? user.discordId ?? ''}
               placeholder="n/a"
-              readOnly
               className="mb-12 lg:mb-20"
               aria-describedby="discord-message"
+              readOnly
+              disabled
             />
 
             <Button type="submit">Save changes</Button>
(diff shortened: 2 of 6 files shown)


app/routes/me.tsx
```diff
@@ -61,64 +69,133 @@ export const action: ActionFunction = async ({request}) => {
...
+        </div>
+      </Grid>
+    </Form>
   )
+
+  // TODO: something... I'm not sure how to fit these into the design
+  //  <Form method="post" action="/me">
+  //    <input type="hidden" name="actionId" value="logout" />
+  //    <button type="submit">Logout</button>
+  //  </Form>
```
smeijer, line 196: That's not true. I got feedback from Gil, and the current position of the logout button is approved. I only still don't entirely know how to adjust it for narrow breakpoints.

app/routes/me.tsx
```diff
@@ -61,64 +69,133 @@ export const action: ActionFunction = async ({request}) => {
...
+  //  Scan this QR code on the other device:
+  //  <img src={data.qrLoginCode} alt="Login QR Code" />
```
smeijer, line 198: Waiting for feedback from Gil

app/routes/me.tsx
```diff
@@ -61,64 +69,133 @@ export const action: ActionFunction = async ({request}) => {
...
+                  href={`https://discord.com/users/${user.discordId}`}
+                >
+                  connected
+                </a>
+              ) : (
+                <a className="text-black dark:text-white" href={authorizeURL}>
+                  Connect to Discord
+                </a>
+              )}
+            </p>
```
smeijer, line 143: This was my own creativity, as the Discord input wasn't included in the design. I think it looks nice with the connect link there on the top right. What do you think?

app/components/icons/logout-icon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
+
+function LogoutIcon() {
+  return (
+    <svg width="24" height="24" fill="none" viewBox="0 0 24 24">
```
smeijer, line 5: TIL that the icons are coming from https://iconic.app. Note to self, replace the other icons as well, as those were copied from figma.

app/routes/me.tsx
```diff
@@ -11,6 +11,14 @@ import {
 } from '../utils/misc'
 import {getMagicLink, updateUser} from '../utils/prisma.server'
 import {requireUser, rootStorage, signOutSession} from '../utils/session.server'
+import {H2, H6} from '../components/typography'
+import {Grid} from '../components/grid'
+import {Input, Label} from '../components/form-elements'
+import {Button} from '../components/button'
+import {CheckIcon} from '../components/icons/check-icon'
+import {TEAM_MAP} from './signup'
```
kentcdodds, line 19: I want to avoid importing/exporting things from route modules. If we need to share, put it in `utils`. When we're all done, I'll come back around and organize things a little bit (hopefully by then we have [this](https://github.com/remix-run/remix/issues/10)).

app/routes/me.tsx
kentcdodds: Yup, I like this. Eventually I think I'll get the user's discord avatar and put it in there somehow. I'll also want a way to disconnect. But I like what you've got here for now 👍

kentcdodds (commented):
Looks excellent so far 👍

app/routes/me.tsx
smeijer: When you add more options, we could add those under a separate section. Just remove this Discord input, and add new headings with all Discord stuff below that. A bit like how `/signup` separates team selection from first name.

app/routes/me.tsx
```diff
@@ -66,57 +74,138 @@ function YouScreen() {
...
-        </details>
-      </div>
+        </Grid>
+      </Form>
+
+      <Grid>
+        <div className="col-span-full mb-12 lg:col-span-5 lg:col-start-8 lg:mb-0">
+          <H2 className="mb-2">Need to login somewhere else?</H2>
+          <H2 variant="secondary" as="p">
+            Scan this QR code on the other device.
```
smeijer, line 197: Still waiting, but added something creative in the mean while. I think it doesn't even look that bad.

app/routes/me.tsx
```diff
@@ -66,57 +74,138 @@ function YouScreen() {
...
+            </Label>
+
+            <Input
+              name="email"
+              id="email-address"
+              autoComplete="email"
+              required
+              defaultValue={user.email}
+              className="mb-8"
+            />
```
smeijer, line 125: Should the email be editable, or `readOnly`?

kentcdodds: Hmmm.... Let's start with readonly. Maybe I'll change it later.

smeijer:
I think this one is ready to get merged. I've marked the email input as being read-only, and Gil has let me know that he's okay with the way I've implemented the QR code. He did mention that it can become a bit large. At the same time, I think that's a good thing, as not all phones have super awesome cameras, and the QR code is one of high density.

That being said, I have created a note on the project board so we can think about it for a moment, and maybe come back to it (and make it smaller). https://github.com/kentcdodds/remix-kentcdodds/projects/1#card-64414167

app/routes/me.tsx
kentcdodds: I like the way it looks. Technically it's challenging though because I don't want the QR code to show up without the user clicking it for two reasons:

1. I want the QR code to expire after a short period
2. I don't want someone else to be able to scan it over the person's shoulder

So it should be hidden until the user clicks something and when they do we'll make a request to generate it.

kentcdodds (commented):
Great progress here 👍

app/routes/me.tsx
smeijer: Makes sense. How about adding an overlay mentioning "click to reveal", exactly where the image is right now. And then when clicked, we fade that overlay out, and the QR in?

kentcdodds: Oh yeah, good idea. And maybe we could put a little 60 second countdown timer in the middle of the QR code. Once the time is up it fades back out and you have to generate a new one. We'll also want a button to hide it earlier. This is all the "fun" and "extra" stuff so if you have other things you want to get done sooner go ahead and just keep this bit out for now.

kentcdodds: Actually, I'll just merge this as-is. We can come back to this later.

PR #32, opened 2021-07-04
fix: support filtering on tags containing spaces

### LINKED ISSUE #31: tag filter on `/blog` doesn't support tags with spaces
The tag filter is a bit buggy, because we split on space while some tags include spaces. Because of that, tags like `open source` cannot be deselected.

https://github.com/kentcdodds/remix-kentcdodds/blob/f1681e5a37456c8af173439ba94bbf13d008ba09/app/routes/blog/index.tsx#L119-L129

smeijer:
fixes #31.


kentcdodds (approved):
Super 👍

PR #41, opened 2021-07-07
feat: implement podcast index & podcast detail page

smeijer:
- Added color & content to the podcast index page
- Added color & content to the podcast detail page
- Created a MDX parser to convert the podcast mdx files to a structured format, that can be rendered in various blocks

TODO:
- The MDX parser coud/should be a remark plugin
- "Homework", should be a list, but I think that's better done via that plugin, as that's where string manipulation gets tricky
- The podcasts don't have a `season` and `episode` property yet, I think that should be extracted from the filename? We could flatten the file structure, and use a name like `S01E02-title.mdx`


app/routes/podcast/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+  console.dir({data})
+  // TODO: save state in url, like the filters on the blog?
+
+  const featured = data.podcasts[0]
+
+  return (
+    <div>
+      <Grid className="grid-rows-max-content mb-36 mt-16">
+        <div className="col-span-full mb-12 lg:col-span-6 lg:col-start-7 lg:row-span-2 lg:mb-0">
+          <img className="object-contain" {...images.kayak()} />
```
kentcdodds, line 76: This is spreading a string (the function returns what you should pass into the `src` prop.

```suggestion
          <img className="object-contain" src={images.kayak()} alt={images.kayak} />
```

smeijer: Got it. And `images.kayak.alt` :+1:

kentcdodds:
FYI, I'm working directly on this PR. Pretty significant changes/improvements to the data structures and content source. Hopefully I'll finish before you wake up 😅

kentcdodds:
It just occurred to me that we actually could probably get lots of this data directly from the simplecast API as well! I'll see what I can come up with. Should hopefully reduce duplication of work for whenever we publish another season.

kentcdodds:
Ok yeah, so we can get a TON (all?) of the stuff we need from the API for the podcast.

I probably won't get this done before you need to work on things more, so could you move on to a different page until I can finish this up?

kentcdodds:
Something to think about is the fact that we actually can even get the raw mp3 path from the API so we could make our own audio player (which we'll probably want for the Call Kent podcast as well). We'll save that for later. I may have someone who can build something like that for us.

smeijer:
> I probably won't get this done before you need to work on things more, so could you move on to a different page until I can finish this up?

Sure! Already working on other pages 😎

kentcdodds:
I'm a bit perplexed by this design and the fact that we have one episode that has multiple guests:

image

Here's the episode with two guests: https://kentcdodds.com/chats-with-kent-podcast/seasons/01/episodes/the-state-of-webassembly-with-lin-clark-and-till-schneidereit

So I'm just trying to figure out how to represent this in the data structure. I'm thinking we'll just need to assume that all episodes can have an array of guests and render them as an array. So then the trick is how do we store the image of the guest. As in that podcast page we just put the two of them together in a single image. I think that's the right way to solve this problem. However, then that guest UI there wouldn't really make sense... I think we can work out something that will look good, but we should keep this in mind...

smeijer:
The picture can indeed be like now, one photo of them together. And you could also define the name like `Lin Clark and Till Schneidereit` as a single string. So that would be a quick fix. But your current page shows the links (Twitter / Github) for both speakers individually, and this design doesn't really offer some space for that. Not sure what to do about that at this moment.

I'll give it some thought.

smeijer:
Regarding all guests as array, that's possible. But as it's the exception to the rule, we could also go with a hybrid. Guest can be an object, or an array of that same object. (`guest: Guest | Guest[]`). Takes a few more lines of code, but might make maintaining the mdx files more comfortable. I think that's where our focus should lie.

kentcdodds:
We're going to not have mdx files when I'm done. We'll get *everything* from the simplecast API which will make this much easier to manage from a content perspective.

kentcdodds:
Not quite ready yet, but it's much closer. We'll definitely want to add mocking and caching for this API because it's pretty slow. But I think we'll be able to manage that without too much trouble. Leave this one to me. I'll get it over the finish line.

kentcdodds:
The social images can (amazingly) be auto-generated by cloudinary: https://support.cloudinary.com/hc/en-us/community/posts/4403508687890-Overlay-capabilities-?page=1

So yeah, this will be completely possible to have all the content come from the API which will be a nice improvement.

smeijer:
> The social images can (amazingly) be auto-generated by cloudinary: https://support.cloudinary.com/hc/en-us/community/posts/4403508687890-Overlay-capabilities-?page=1

Oh, that's nice! I did not expect that.

kentcdodds:
Alrighty, I've got mocking in place so we can work on this before we actually have all the podcast metadata up-to-date. Hopefully that'll get up-to-date soon, but I like the idea of mocking every third party API so we can work completely offline and remain reliable during development anyway so this works nicely.

Still need to add: 

- [ ] Support for route-based tabs per season `/chats/:seasonNumber/`
- [ ] Redis caching
- [ ] Support for fetching individual episode rather than every episode

This should all be done today I think.

PR #48, opened 2021-07-14
fix: resolve all heading order violations

smeijer:
Removed level skipping in headings, and removed some headings altogether. I've also added headings visible to screen-readers only above the guest listing on the podcast pages, and above the courses on the courses page.


kentcdodds:
Awesome 👍

PR #50, opened 2021-07-14
feat: add 500 page

smeijer:
Added a server error page that can be rendered from the ErrorBoundary. A simple copy of the 404, but with different captions.

### DIFF
--- a/app/components/sections/hero-section.tsx
+++ b/app/components/sections/hero-section.tsx
@@ -79,7 +79,7 @@ function HeroSection({
         )}
       >
         <div className="flex flex-auto flex-col">
-          <H2 as="p">{title}</H2>
+          <H2 as="h2">{title}</H2>
           {subtitle ? (
             <H2 as="p" variant="secondary" className="mt-3">
               {subtitle}
--- a/app/routes/404.tsx
+++ b/app/routes/404.tsx
@@ -1,13 +1,15 @@
 import * as React from 'react'
+import {ErrorPage} from '../components/errors'
 
 export function meta() {
   return {title: "Ain't nothing here"}
 }
 
-export default function FourOhFour() {
+export default function NotFoundPage() {
   return (
-    <div>
-      <h2>404</h2>
-    </div>
+    <ErrorPage
+      title="404 - Oh no, how did you get here?"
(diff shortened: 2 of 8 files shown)


app/utils/mdx.tsx
```diff
@@ -233,13 +233,41 @@ function FourOhFour() {
   )
 }
 
+function ServerError() {
```
kentcdodds, line 236: I'm not sure the `utils/mdx.tsx` util is the best place for this. Maybe `utils/misc.tsx`?

smeijer: Yeah, I just placed it next to the 404. How do you feel about moving them both to `components/errors.tsx`?

kentcdodds: Yeah, let's do 👍

kentcdodds: The 404 component in there was originally intended for blog posts and other mdx-based pages. But I guess it could be used for other things now too.

app/routes/404.tsx
```diff
@@ -1,13 +1,15 @@
...
-export default function FourOhFour() {
+export default function NotFoundPage() {
   return (
-    <div>
-      <h2>404</h2>
-    </div>
+    <ErrorPage
+      title="404 - Oh no, how did you get here?"
+      subtitle="This is not a page on kentcdodds.com. So sorry."
+    />
```
smeijer, line 13: This page seems to be a bit special to remix. If I would render `<FourOhFour>` here directly, it throws:

```shell
    You must render this element inside a <Remix> element

      30 |
      31 | function FourOhFour() {
    > 32 |   const matches = useMatches()
         |                   ^
      33 |   const last = matches[matches.length - 1]
      34 |   const pathname = last?.pathname
```

app/components/sections/hero-section.tsx
```diff
@@ -79,7 +79,7 @@ function HeroSection({
         )}
       >
         <div className="flex flex-auto flex-col">
-          <H2 as="p">{title}</H2>
+          <H2 as="h2">{title}</H2>
```
smeijer, line 82: Revert this to a heading. It's the title of the page. I'm not sure what I was thinking before.

kentcdodds (approved):
Super 👍

PR #52, opened 2021-07-15
fix: update footer links

smeijer:
Pointed the footer links to existing pages where possible. I also added `credits`, and noticed that `Discord` is listed twice.

### DIFF
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -63,7 +63,7 @@ function ContactSection() {
       <ul className="mt-4">
         <FooterLink name="contact page" href="/contact" />
         <FooterLink name="office hours" href="/office-hours" />
-        <FooterLink name="Join discord" href="https://kcd.im/discord" />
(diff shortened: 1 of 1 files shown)


app/components/footer.tsx
```diff
@@ -63,7 +63,7 @@ function ContactSection() {
       <ul className="mt-4">
         <FooterLink name="contact page" href="/contact" />
         <FooterLink name="office hours" href="/office-hours" />
-        <FooterLink name="Join discord" href="https://kcd.im/discord" />
+        <FooterLink name="Join discord" href="/discord" />
```
smeijer, line 66: Discord is also mentioned under `sitemap`. Should it be listed twice?

kentcdodds: Yeah, let's replace this with:

```suggestion
        <FooterLink name="Call Kent" href="/call" />
```

kentcdodds (approved):
Super 👍

PR #54, opened 2021-07-16
feat: animate arrow on course card buttons

smeijer:
https://user-images.githubusercontent.com/1196524/125960607-43a931fb-9cca-4db6-8203-b8c67f1aa6be.mp4

### DIFF
--- a/app/components/course-card.tsx
+++ b/app/components/course-card.tsx
@@ -1,8 +1,16 @@
 import * as React from 'react'
+import {motion, Variants} from 'framer-motion'
 import {ArrowIcon} from './icons/arrow-icon'
 import {H2} from './typography'
 import {ButtonLink} from './button'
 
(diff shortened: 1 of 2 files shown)


app/components/course-card.tsx
```diff
@@ -1,8 +1,16 @@
 import * as React from 'react'
+import {motion, Variants} from 'framer-motion'
 import {ArrowIcon} from './icons/arrow-icon'
 import {H2} from './typography'
 import {ButtonLink} from './button'
 
+const MotionButtonLink = motion(ButtonLink)
```
smeijer, line 7: I've added the animation stuff in this component, instead of in `Button`, because this seems to be the only case of this type of button. 

If we'd made the abstraction in `Button`, we get something as complicated as `ArrowLink`.

kentcdodds (approved):
Super!

PR #55, opened 2021-07-16
feat: add redbox for server errors in dev env

smeijer:
Improved the 500-error page by showing the error when in development mode. It's dismissable by clicking on the backdrop. 

I haven't tested, but due to the check on `NODE_ENV` it shouldn't appear in production mode.

https://user-images.githubusercontent.com/1196524/125967843-8e9e01d2-6670-4fc4-a995-0676defa4854.mp4


kentcdodds (approved):
Thanks 👍

PR #56, opened 2021-07-16
fix: squishing of podcast prev/next buttons

smeijer:
Fixed the alignment of the prev/next buttons below the podcast player on mobile screens, and the squashing of the images in them.


kentcdodds (approved):
👍

Issue #53, opened 2021-07-16
add permlink to frontmatter?


smeijer:
Copy buttons need a full URL to be copied to the clipboard. 

I can construct something using `new URL(frontmatter.slug, location.origin).toString()`, but this will not work server-side. 

My proposal; add a `BASE_URL` env var, and use that to construct perma-links that will be part of the frontmatter.

kentcdodds:
Was about to write this but then decided to just make it work anyway and it's not all that bad. Here's what I was going to write:

We can't use `BASE_URL` as a global env variable because we don't know for sure what the domain URL will be. We could just hard code it to `kent.dev`, but then when we move over to `kentcdodds.com` or if I ever do deploy previews all the links will be wrong. That's why I created:

https://github.com/kentcdodds/remix-kentcdodds/blob/930308323f4c0182b5ca57d14a7d9a27f54b1cc0/app/utils/misc.tsx#L162-L170

We use that to put the `origin` in the `requestInfo` in our root route:

https://github.com/kentcdodds/remix-kentcdodds/blob/930308323f4c0182b5ca57d14a7d9a27f54b1cc0/app/root.tsx#L103-L107

Then that's put into context:

https://github.com/kentcdodds/remix-kentcdodds/blob/930308323f4c0182b5ca57d14a7d9a27f54b1cc0/app/root.tsx#L176

And we use that to create things like the discord authorization URL: https://github.com/kentcdodds/remix-kentcdodds/blob/adb6eb3ff59b4678a65b49547dc6aceffa897249/app/routes/me.tsx#L80

So we could do the same for permalinks.

I tried adding a `permalink` property to the `MdxPage` type, but that just made things much more complicated because we had to pass the request object (or at least the `origin` everywhere we created `MdxPage` objects) ... hold up. Let me try a different approach... Yeah, that's not all that bad. `MdxPage` now has a `permalink` property.

kentcdodds:
Actually, I changed my mind again. We don't need the permalink all that often, so I think I'd prefer to just construct it wherever we need it:

```tsx
const requestInfo = useRequestInfo()
const permalink = `${requestInfo.origin}/blog/${page.slug}`
```

smeijer:
`useRequestInfo().origin` looks great! Let's use that.

PR #57, opened 2021-07-16
feat: improve copy button

smeijer:
Improved the copy-button:

- added copy functionality
- added a "copy confirmation" state
- moved the button on the featured section inside the photo (match the gallery cards)
- removed the copy button from all other feature sections

**TODO:**
- set article urls to be copied (blocked by #53 / decision making)

**PREVIEW:**

https://user-images.githubusercontent.com/1196524/125971669-1256f429-518e-4d6c-a41c-874a18d0c8f8.mp4

### DIFF
--- a/app/routes/404.tsx
+++ b/app/routes/404.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
-import {ErrorPage} from '../components/errors'
+import {HeroSection} from '../components/sections/hero-section'
+import {images} from '../images'
 
 export function meta() {
   return {title: "Ain't nothing here"}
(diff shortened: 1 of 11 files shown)


app/routes/404.tsx
```diff
@@ -1,15 +1,20 @@
...
   return {title: "Ain't nothing here"}
 }
 
 export default function NotFoundPage() {
   return (
-    <ErrorPage
-      title="404 - Oh no, how did you get here?"
-      subtitle="This is not a page on kentcdodds.com. So sorry."
-    />
+    <main>
```
smeijer, line 11: This was a bit unexpected, but when I use the `useRequestInfo` in the 404 (via error > article recommendations), an error was thrown mentioning that the component needs to be wrapped in `RequestInfoProvider`. After adding that, I ended up with an error mentioning that I need to run in a `remix route element`.

I didn't think it's worth going down this rabbit hole, so I replaced the root 404 with something simpler.

kentcdodds: Yeah, I actually ran into a similar problem when I tried to remove the article fixtures. I'll look into it later. Thanks!

PR #63, opened 2021-07-18
feat: make translations look less bad

### LINKED ISSUE #60: Make translations look not bad
Check https://kent.dev/blog/concerning-toolkits for an example of how it looks when an author is included.

Check https://kent.dev/blog/application-state-management-with-react for an example of how it looks when there are multiple.

smeijer:
An attempt to fix #60.

I've also tried adding the up-right arrow to the pills, but with multiple translations that gets quite crowded.

https://user-images.githubusercontent.com/1196524/126079729-94644abd-1de1-43f5-97d1-23d7e3cbdc64.mp4


kentcdodds (approved):
👍👍

PR #65, opened 2021-07-19
fix: limit height of hero for ultra-high screens

### LINKED ISSUE #64: [Suggestion] Reduce spacing of head elements at 4k resolution
Hi all,

sorry to chime in here, but I would recommend reducing the spacing on large resolutions. For illustration see the attached screenshots.

image
image

smeijer:
fixes #64 

Maybe we should also limit the max-width of the navbar to stay closer to the content width? A bit wider as content, but not as wide as shown in #64?

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -85,7 +85,7 @@ module.exports = {
         '10vw': '10vw', // page margin
       },
       height: {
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks 👍

kentcdodds:
Yeah, and the navbar thing would be good too

PR #67, opened 2021-07-19
feat: add team read count to blog

smeijer:
I've added the team stats to the blog pages. 

https://user-images.githubusercontent.com/1196524/126185154-116d182e-f393-4d1b-8c6c-1bb2378d1180.mp4

### DIFF
--- a/app/components/team-stats.tsx
+++ b/app/components/team-stats.tsx
@@ -3,7 +3,7 @@ import {motion} from 'framer-motion'
 import clsx from 'clsx'
 import type {Team} from '@prisma/client'
 
-const dotColors: Record<Team, string> = {
+const rankingColors: Record<Team, string> = {
   RED: 'bg-team-red',
   YELLOW: 'bg-team-yellow',
   BLUE: 'bg-team-blue',
@@ -35,7 +35,7 @@ function Stat({totalReads, team, percent}: ReadRanking) {
         }}
         className={clsx(
           'z-10 flex items-center justify-end px-2 bg-black rounded-l-md origin-right',
-          dotColors[team],
+          rankingColors[team],
         )}
       >
         <motion.span
--- a/app/components/team-stats.tsx
+++ b/app/components/team-stats.tsx
@@ -1,8 +1,6 @@
 import * as React from 'react'
 import {motion} from 'framer-motion'
 import clsx from 'clsx'
-import {useOptionalUserInfo, useTeam} from '../utils/providers'
-import {alexProfiles} from '../images'
 import type {Team} from '@prisma/client'
 
 const dotColors: Record<Team, string> = {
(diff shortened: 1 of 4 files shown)


kentcdodds:
Oh sweet!

So here's what I was thinking with this. I want it to be a little competitive, so the way it's displayed should show which team is "winning" without having to compare numbers. I've thought about doing a pie chart or bar charts or something.

We can show the total number of reads from each team, but to make it fair (because people can choose their own teams), the actual numbers in the comparison would be the percentage of active members of a team who have read the post in the last 6 months. "Active member" is defined as someone who has read a post in the last year.

Hopefully that makes sense. What do you think would be the best way to display that?

Also, I've considered theming the post after the team that's winning. So a slight yellow tint to the background color if the yellow team is leading for example.

smeijer:
Yeah, makes total sense.

I'm not sure about the pie chart. Maybe simply because I can't picture it. Where it will be placed, and how it behaves. 

The bar chart on the other hand. Replace the dots with right aligned horizontal bars. And make them grow on hover, while showing the label. The idea like in the video, but with right aligned bars. 

Alternatively, we could do the same thing as in the video above, but scale the dots depending on score/ranking. We'd loose the "logged in state" as that dot already renders larger, messing up the relative size differences. But gain a clean way of showing the best performing team. 

Bar width or dot size, could either be set to 3 clearly distinctive sizes or be determined based on the scores. The former might show more clearly who's in the lead, while the latter acts more as a chart, highlighting close calls.

I think the 3 fixed sizes is best, as I'd expect that due to the normalization team scores will mostly be quite close to each other.

kentcdodds:
What if we do right-aligned bars where the first place and last place have a set size, but 2nd place will be set to a size relative to the 2nd team's percentage (so if it's close to 1st place, it'll be closer in size to 1st place than 3rd)?

And we could possibly put the avatar to the right of the bar of their team. I like keeping that avatar if we can.

smeijer:
Yeah, something like that. I'll fix something.

kentcdodds:
I'm working on getting the data for you. The loader will give you this:

```
type LoaderData = {
  page: MdxPage | null
  recommendations: Array<MdxListItem>
  readRankings: Record<Team, number>,
}
```

kentcdodds:
I'm also working on getting the prisma seed script to auto-add some of this data so we don't have to fake it out locally or anything.

kentcdodds:
We'll also want to think about how to deal with a tie for 1st or 2nd. Probably just make them the same size.

smeijer:
A shared first place results in two equality sized wide bars. And a shared second place, should result in two equality sized small bars (same as 3th place).

kentcdodds:
Actually, I've changed it a bit. You need 3 numbers:

1. Total number of reads by a given team
2. Total reads in the last 6 months
3. Total active team members

It's possible (but unlikely) that total reads could be higher than active team members. But we'll rank folks based on `Total Reads` / `Total Active Members` and on hover we'll display the total reads by a given team.

It's possible that one team will have a higher total read count, but still be in 2nd or 3rd place. For this reason (and others), we should probably have a little question mark icon that links to a page that I'll write that will explain the ranking system.

kentcdodds:
Ok, here's where I finally settled:

```ts
type LoaderData = {
  page: MdxPage | null
  recommendations: Array<MdxListItem>
  readRankings: Array<{team: Team; totalReads: number; percent: number}>
}
```

I figured there wasn't any use in forcing you to make the calculation on the client. I'll just send the percent the bar should be set to. I'm thinking the rankings should be sorted highest to lowest (to avoid giving preferential treatment), so I've sorted them for you on the backend. You should be able to just iterate directly over the `data.readRankings` and render the bars.

I've updated the seeder to add some reads for `http://localhost:3000/blog/super-simple-start-to-remix`, so feel free to use that post to test this with. Just run `npx prisma migrate reset --force` to reset your database and seed it with the test data.

smeijer:
This is what I came up with. The vertical line renders in the `current-team-color`, and the bars are sorted (top > down) `[last, first, second]`. Just like a stage, but turned 90 degrees. I think that looks better than a stair like structures. 

I've added a screenshot for comparison, note that the three variants are based on the same stats. It's purely the order that has changed.

I've also tried adding the avatar to the right of these bars, as you suggested. But it looks crowded/forced that way. I'll try to come up with some other ideas for that.

https://user-images.githubusercontent.com/1196524/126294662-3a267ec8-9b18-47ee-b4be-feba8568819c.mp4


image

kentcdodds:
Looking great! I think we could make them a bit wider to make the difference more obvious. If we can't make the avatar work, that's fine.

app/components/team-stats.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react'
+import {motion} from 'framer-motion'
+import clsx from 'clsx'
+import type {Team} from '@prisma/client'
+
+const dotColors: Record<Team, string> = {
```
kentcdodds, line 6: This should probably be `barColors` or `rankingColors`

app/components/team-stats.tsx
```diff
@@ -0,0 +1,74 @@
...
+        )}
+      >
+        <motion.span
+          variants={{
+            initial: {opacity: 0, scale: 0.5},
+            hover: {opacity: 1, scale: 1},
+          }}
+          className="text-gray-900 text-lg font-medium"
+        >
+          {new Intl.NumberFormat().format(totalReads)}
```
kentcdodds, line 48: Nice! I wouldn't have thought of formatting it like this. Thanks 👍

kentcdodds:
Decided it looks great so I'm just going to merge 👍

Thanks!

smeijer:
I'll submit a new PR for the fixes :+1:

PR #71, opened 2021-07-20
feat: render team-charts horizontally and fixed at top/bottom of article

smeijer:
- increased the size difference between bars
- rendered them horizontally
- fixed to top and bottom of the article, instead of sticky on screen
- added user avatar in case they're logged in

https://user-images.githubusercontent.com/1196524/126388598-facfe787-e064-44df-a2ed-6bd24023c1d5.mp4

### DIFF
--- a/app/components/team-stats.tsx
+++ b/app/components/team-stats.tsx
@@ -17,40 +17,60 @@ type ReadRanking = {
   percent: number
 }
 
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Excellent! Great use of the `direction` prop 👍

PR #72, opened 2021-07-20
Switch to docker for builds

mrkurt:


smeijer:
Big fan of the commit messages! 😂

smeijer:
I've allowed your actions to run, sorry for the outcome.

smeijer:
From now on, the actions should run automatically.

kentcdodds:
Looks like we kinda need this before this makes much of a significant improvement: https://github.com/docker/build-push-action/issues/417

mrkurt:
This appears to be working pretty well now, the "Docker Build" step on the last run in my repo took <1 min and used cached layers.

kentcdodds:
How fast will that deploy?

mrkurt:
Probably ~2 minutes. The resulting image is 70% smaller so it'll be faster to pull, we do a rolling deploy and update each of your VMs one by one, so total deploy time is a function of `count * (image-pull-time + time-to-healthy)`.

kentcdodds:
Thanks a million Kurt!

PR #75, opened 2021-07-21
feat: implement call page

smeijer:
Added color and content to the `/call` page. The design is inspired by `/chats` (podcast), but instead of opening every call in a details page, they expand and show the player inline.

**TODOS**

- Alex is rendered, but in low quality. We need higher resolution images

**IDEAS**

_copied from Discord:_

- Add search (like on blog page)
- Add keyword filter (like on blog page)
- Display summary on each row, and the description in expanded view?
- Display the keywords on each row as well?

### DIFF
--- a/app/routes/call.tsx
+++ b/app/routes/call.tsx
@@ -91,7 +91,7 @@ export default function CallHomeScreen() {
       />
 
       {/*
-        IDEA: when there will be manyu episodes, we could split this by year, and
+        IDEA: when there will be many episodes, we could split this by year, and
         display it with tabs like on the podcast page. [2023, 2022, 2021]
       */}
       <Grid className="mb-24 lg:mb-64">
--- a/app/routes/call.tsx
+++ b/app/routes/call.tsx
@@ -126,84 +126,86 @@ export default function CallHomeScreen() {
         </div>
 
         <div className="col-span-full">
-          {sortedEpisodes.map((episode, idx, {length}) => {
-            const number =
-              sortOrder === 'asc' ? idx + 1 : Math.abs(length - idx)
+          <CallKentEpisodesProvider value={data.episodes}>
+            {sortedEpisodes.map((episode, idx, {length}) => {
+              const number =
+                sortOrder === 'asc' ? idx + 1 : Math.abs(length - idx)
 
-            return (
-              <div
-                className="border-b border-gray-200 dark:border-gray-600"
-                key={episode.slug}
-              >
-                <Link
-                  data-episode={episode.slug}
+              return (
+                <div
+                  className="border-b border-gray-200 dark:border-gray-600"
                   key={episode.slug}
-                  to={activeSlug === episode.slug ? './' : `./${episode.slug}`}
                 >
-                  <Grid nested className="group relative py-10 lg:py-5">
-                    <div className="bg-secondary absolute -inset-px group-hover:block hidden -mx-6 rounded-lg" />
-                    <div className="relative flex-none col-span-1">
-                      <div className="absolute inset-0 flex items-center justify-center opacity-0 group-hover:opacity-100 transform scale-0 group-hover:scale-100 transition">
-                        <div className="flex-none p-4 text-gray-800 bg-white rounded-full">
-                          <TriangleIcon size={12} />
+                  <Link
+                    data-episode={episode.slug}
+                    key={episode.slug}
+                    to={
+                      activeSlug === episode.slug ? './' : `./${episode.slug}`
+                    }
+                  >
+                    <Grid nested className="group relative py-10 lg:py-5">
+                      <div className="bg-secondary absolute -inset-px group-hover:block hidden -mx-6 rounded-lg" />
+                      <div className="relative flex-none col-span-1">
+                        <div className="absolute inset-0 flex items-center justify-center opacity-0 group-hover:opacity-100 transform scale-0 group-hover:scale-100 transition">
+                          <div className="flex-none p-4 text-gray-800 bg-white rounded-full">
+                            <TriangleIcon size={12} />
+                          </div>
+                        </div>
+                        <img
+                          className="w-full rounded-lg object-cover"
+                          src={episode.imageUrl}
+                          alt={episode.title}
+                        />
+                      </div>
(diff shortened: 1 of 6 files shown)


app/routes/call/$slug.tsx
```diff
@@ -24,12 +25,17 @@ export default function Screen() {
...
+
   return (
-    <div>
-      <pre>{JSON.stringify(episode, null, 2)}</pre>
-    </div>
+    <div
+      dangerouslySetInnerHTML={{
+        __html: theme === 'dark' ? episode.embedHtmlDark : episode.embedHtml,
+      }}
+    />
```
smeijer, line 39: Note that this doesn't work with fake data, as the iframe then doesn't load. Replace with the below for testing purpose:

`<iframe src='https://share.transistor.fm/e/1493e91f/dark' width='100%' height='180' frameborder='0' scrolling='no' seamless='true' style='width:100%; height:180px;'></iframe>`.

app/routes/call.tsx
```diff
@@ -20,30 +34,183 @@ export const loader: LoaderFunction = async ({request}) => {
...
+    sortOrder === 'desc' ? data.episodes : [...data.episodes].reverse()
+
+  // An effect to scroll to the episode's position when opening a direct link
+  React.useEffect(() => {
+    if (activeSlug) {
+      document.querySelector(`[data-episode="${activeSlug}"]`)?.scrollIntoView()
+    }
+    // we really only want to run this on mount. Otherwise it will hijack the
+    // users scroll while they browse the episodes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kentcdodds, line 68: In this case we can use a ref so we don't have to disable the rule

app/routes/call/$slug.tsx
kentcdodds: We could update the mocks in `mocks/transistor.ts`

app/routes/call.tsx
```diff
@@ -20,30 +34,183 @@ export const loader: LoaderFunction = async ({request}) => {
...
+                            opacity: 1,
+                          },
+                        }}
+                        initial="collapsed"
+                        animate="expanded"
+                        exit="collapsed"
+                        transition={{duration: 0.15}}
+                        className="relative col-span-full"
+                      >
+                        <CallKentEpisodesProvider value={data.episodes}>
```
kentcdodds, line 196: Maybe we should move this out to wrap the whole screen so we don't have to render it for every episode

app/routes/call.tsx
```diff
@@ -20,30 +34,183 @@ export const loader: LoaderFunction = async ({request}) => {
...
+              <>
+                Showing newest first
+                <ChevronDownIcon className="ml-2 text-gray-400" />
+              </>
+            )}
+          </button>
+        </div>
+
+        <div className="col-span-full">
+          {sortedEpisodes.map((episode, idx, {length}) => {
```
kentcdodds, line 128: Should we try and use `@reach/accordion` for this so we don't have to worry about making sure it's accessible?

app/routes/call.tsx
```diff
@@ -20,30 +34,183 @@ export const loader: LoaderFunction = async ({request}) => {
...
+                      </div>
+                      <img
+                        className="w-full rounded-lg object-cover"
+                        src={episode.imageUrl}
+                        alt={episode.title}
+                      />
+                    </div>
+                    <div className="text-primary relative flex flex-col col-span-3 md:col-span-7 lg:flex-row lg:col-span-11 lg:items-center lg:justify-between">
+                      <div className="mb-3 text-xl font-medium lg:mb-0">
+                        {/* TODO: make it three digits? How many calls do we expect? */}
```
kentcdodds, line 158: Yeah, let's be optimistic and do three digits 😅

app/routes/call.tsx
```diff
@@ -20,30 +34,183 @@ export const loader: LoaderFunction = async ({request}) => {
...
+                        <span className="inline-block w-10 lg:text-lg">
+                          {`${number.toString().padStart(2, '0')}.`}
+                        </span>
+
+                        {episode.title}
```
kentcdodds, line 163: Could/should we also display the episode summary and keywords in a small font here? I'm not sure about the summary, but keywords might be useful. The problem is, I'm not sure what the summary will be. Right now we don't connect that information anyway...

kentcdodds (commented):
Awesome! Man I'm really amazed. Just a few things

app/routes/call.tsx
smeijer: Makes sense. Thanks!

app/routes/call.tsx
smeijer: I've made them dynamic. :sunglasses: It now shows as many leading zeros as required, based on the number of episodes available. You're good until episode 9999. Probably that's the right time to add some windowing or pagination as well :sweat_smile:

kentcdodds: I like your idea of splitting things by year 👍 We'll see how things look if/when we get to a ton of episodes :)

app/routes/call.tsx
smeijer: With summary you mean `description` right? There is no limit on how long that can get. That's why I showed it inside the accordion.

I'll add the keywords. But there also, there is no limit on how many. Should we stick to the first 3? Separated by space?

app/routes/call.tsx
smeijer: I'm not sure. It's just a `link` that navigates to a new page. Does it make sense to use accordion for that? We use it in `/discord`, but the accordions there don't navigate. Does that make a difference?

Happy to add it. Just wanna make sure that there is value in it.

app/routes/call.tsx
smeijer: Maybe even just add both in the accordion. Looks quite nice!

image

kentcdodds: Why don't we put the keywords (unlabeled) in muted text below the title? I just want to make it easy to scan the episodes. Maybe that's not necessary though 🤔

app/routes/call.tsx
kentcdodds: As long as it's keyboard accessible then I think we're good.

app/routes/call.tsx
smeijer: I've created #77. Let's take another look at that asap.

Issue #69, opened 2021-07-20
force aspect ratio of blog header images?


smeijer:
Should we force the aspect ratio, or at least max-height of blog header images, so that the image of articles like `http://localhost:3000/blog/super-simple-start-to-remix` don't span more than my viewport height?

kentcdodds:
Yes

PR #76, opened 2021-07-21
feat: render content/pages using prose classes

smeijer:
I've applied the same header/image/content logic to the content pages, as we do for blog articles. This results in pretty rendering for:

- /appearances
- /clubs
- /conduct
- /info
- /links
- /transparency

I think the pages would benefit immensely if we'd added some banner images to them.

### DIFF
--- a/app/routes/$slug.tsx
+++ b/app/routes/$slug.tsx
@@ -9,6 +9,9 @@ import {
 } from '../utils/mdx'
 import {FourOhFour} from '../components/errors'
 import {getUser} from '../utils/session.server'
+import {Grid} from '../components/grid'
+import {BackLink} from '../components/arrow-button'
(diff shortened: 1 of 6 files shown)


kentcdodds (approved):
Awesome! 

By banner, you mean like in blog posts?

smeijer:
> By banner, you mean like in blog posts?

Exactly.

kentcdodds:
Alright, do you want to add support for banners for the relevant pages and I'll add a banner to the mdx 👍

smeijer:
I already did. It should work when you add `bannerUrl` and `bannerAlt` to the frontmatter.

I only need to fix the aspect ratio, I'll do that as part of #69.

kentcdodds:
Oh, fantastic 👍 Thanks :)

PR #78, opened 2021-07-21
feat: make call recorder look less terrible

smeijer:
Added "all the" call recorder pages. They ain't awesome. But it's a start.


smeijer:
I think I should render the keywords here as well, just as in #75

kentcdodds (approved):
LET'S DO IT! Thanks 👏

PR #81, opened 2021-07-22
fix: use native anchors for external links

smeijer:
Add support for external links to the `ButtonLink`, which is used by the `CourseCard`. Without this, the destination url becomes `https://kent.dev/https://testingjavascript.com`.


kentcdodds (approved):
Awesome 👍

PR #82, opened 2021-07-22
feat: add talks page

smeijer:
Created the talks page.

### DIFF
--- a/app/routes/talks.tsx
+++ b/app/routes/talks.tsx
@@ -6,6 +6,7 @@ import type {Await} from 'types'
 import {useRef, useState} from 'react'
 import formatDate from 'date-fns/format'
 import {Link, useLocation} from 'react-router-dom'
+import clsx from 'clsx'
 import slugify from '@sindresorhus/slugify'
 import {typedBoolean} from '../utils/misc'
 import {markdownToHtml} from '../utils/markdown.server'
@@ -156,7 +157,7 @@ function Card({
 
         <div className="flex space-x-2">
           {tag ? (
-            <div className="inline-block -mr-8 -mt-4 mb-4 px-8 py-4 text-black dark:text-white text-lg dark:bg-gray-600 bg-white rounded-full">
+            <div className="inline-block -mr-8 -my-4 px-8 py-4 text-black dark:text-white text-lg dark:bg-gray-600 bg-white rounded-full">
               {tag}
             </div>
           ) : null}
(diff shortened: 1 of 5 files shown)


app/routes/talks.tsx
```diff
@@ -87,15 +97,182 @@ export const loader: LoaderFunction = async ({request}) => {
...
+        <div className="flex space-x-2">
+          {tag ? (
+            <div className="inline-block -mr-8 -mt-4 mb-4 px-8 py-4 text-black dark:text-white text-lg dark:bg-gray-600 bg-white rounded-full">
+              {tag}
+            </div>
+          ) : null}
+        </div>
+      </div>
+
+      <H3 as="div" className="flex flex-none items-end mb-4 h-48">
```
kentcdodds, line 146: Let's add `import slugify from '@sindresorhus/slugify'` to the top and then above: `const slug = slugify(title)` and then:

```suggestion
      <H3 as="a" href={`#${slug}`} id={slug} className="flex flex-none items-end mb-4 h-48">
```

Or something like that. So we can link directly to one of the talks. I don't know whether it would be possible to somehow highlight the one that's selected as well, but that would probably be good too. I link people to my talks pretty often.

app/routes/talks.tsx
```diff
@@ -87,15 +97,182 @@ export const loader: LoaderFunction = async ({request}) => {
...
+}: LoaderData['talks'][0]) {
+  const latestDate = deliveries
+    .filter(x => x.date)
+    .map(x => new Date(x.date as string))
+    .sort((l, r) => r.getTime() - l.getTime())[0] as Date
+
+  const isInFuture = latestDate.getTime() > Date.now()
+
+  // `software development` makes a terrible tag to render in this position
+  const tag = tags.filter(x => x.length <= 10)[0]
```
kentcdodds, line 121: Maybe we could sort all the tags (for all the talks) by the most frequently used and then choose the tag of this talk that is the most frequently used by the talks altogether and use that one.

kentcdodds (commented):
Looking great!

app/routes/talks.tsx
smeijer: The highlight isn't an issue, but the slug is. Talk titles agent unique due to the TBA's.

app/routes/talks.tsx
smeijer: Yeah, that makes sense. Should we also use that to limit the number of options in the filter? Only show tags that are used `n` times? (need to find a nice number)

app/routes/talks.tsx
smeijer: Oh, sweet. This will work.

```js
import {slugifyWithCounter} from '@sindresorhus/slugify';

const slugify = slugifyWithCounter();

slugify('foo bar');
//=> 'foo-bar'

slugify('foo bar');
//=> 'foo-bar-2'

slugify.reset();

slugify('foo bar');
//=> 'foo-bar'
```

app/routes/talks.tsx
kentcdodds: Yup. We could just limit it to a total number of tags. Maybe the top 15-20?

smeijer: I'm now showing every tag that's being used by at least 2 talks. That results in 9 filterable tags.

smeijer:
This is ready if you are. Let's fix #85 another day.

dodas:
Hey @kentcdodds , I was watching the office hours stream where you were talking about how the "routing" is implemented on this page. However, I noticed that server navigation to specific talks doesn't work and responds with `null`:
https://kentcdodds.com/talks/don-t-solve-problems-eliminate-them

smeijer:
@dodas , that definitely deserves an issue ticket so it doesn't get forgotten. Do you want to create that?

kentcdodds:
I know what's going on. This broke when resource routes were introduced. We just need to have a default export of a component that renders nothing.

PR #83, opened 2021-07-22
Docker deploy (round 2)

mrkurt:
This fixes the previous docker build issues, builds + pushes the image in one step, and then deploys the image separately. It's a pretty nice speedup!

image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 node_modules
+Dockerfile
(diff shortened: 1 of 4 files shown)


mrkurt:
The build portion of the pipeline needs a `FLY_API_TOKEN` to do its thing, I think GitHub is preventing my run from using that token for security reasons.

I'm splitting the build/push steps so peoples' PRs don't break the build.

kentcdodds:
Ok, here we go!!!!

PR #90, opened 2021-07-24
feat: add caption above talks filter

smeijer:
Added a missing caption so that the filter has the same visual appearance as on `/blog`.

image


kentcdodds (approved):
🔎

PR #91, opened 2021-07-25
feat: force aspect ratio on blog banner images

### LINKED ISSUE #69: force aspect ratio of blog header images?

smeijer:
I've forced the aspect ratio to 3/4 (vertical) on small screen and 3/2 (horizontal) on medium and up.

fixes #69


kentcdodds (approved):
Sweet 👍

PR #95, opened 2021-07-27
feat: add animated team ring

smeijer:
A video is worth a thousand words

https://user-images.githubusercontent.com/1196524/127164580-e2fdb775-c46a-4c18-baed-7df5fa864fe1.mp4

### DIFF
--- a/app/components/hooks/use-element-state.tsx
+++ b/app/components/hooks/use-element-state.tsx
@@ -1,0 +1,86 @@
+import {useRef, useEffect, useState, RefCallback, useCallback} from 'react'
+
+type State = 'active' | 'focus' | 'hover' | 'initial'
+
+// This started as a work around for https://github.com/framer/motion/issues/1221,
+// but it's so much more now. The variants in framer motion support hover, focus
+// and tap, while this effect also listens to the keypress, so that `Enter`
+// results in an active state as well.
+function useElementState(): [RefCallback<HTMLElement>, State] {
+  const ref = useRef<HTMLElement | null>(null)
+  const [state, setState] = useState({
+    focus: false,
+    hover: false,
+    active: false,
+  })
+
+  const setRef: RefCallback<HTMLElement> = useCallback(element => {
+    ref.current = element
+  }, [])
+
+  useEffect(() => {
+    const el = ref.current
+    if (!el) return
+
+    const pointerenter = () => setState(s => ({...s, hover: true}))
+    const pointerleave = () => setState(s => ({...s, hover: false}))
+    const focus = () => setState(s => ({...s, focus: true}))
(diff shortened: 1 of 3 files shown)


app/components/navbar.tsx
```diff
@@ -235,6 +239,78 @@ function MobileMenu() {
...
 }
 
+// Timing durations used to control the speed of the team ring in the profile button.
+// Time is seconds per full rotation
+const durations = {
+  initial: 20,
+  hover: 10,
+  focus: 4,
+  active: 0.25,
+}
```
smeijer, line 249: You can play a bit with these to get speeds that look good to you. The lower the value, the faster it rotates.

app/components/navbar.tsx
```diff
@@ -235,6 +239,78 @@ function MobileMenu() {
...
+      )}
+      ref={ref}
+    >
+      <motion.div className="absolute" animate={controls}>
+        <TeamCircle size={56} team={team} />
+      </motion.div>
+      <img
+        className={clsx('inline w-10 h-10 rounded-full select-none', {
+          'object-cover bg-inverse': !isMascotAvatar,
+          'object-contain': isMascotAvatar,
```
smeijer, line 305: I've changed these so that Alex looks awesome! For normal photos, we want them to "cover" the image circle. Alex on the other hand is an illustration with a transparent background.

app/components/team-circle.tsx
```diff
@@ -0,0 +1,123 @@
...
+      </svg>
+    )
+  }
+
+  const [teamOne, teamTwo] = Object.keys(colors).filter(x => x !== team) as [
+    OptionalTeam,
+    OptionalTeam,
+  ]
+
+  const teamSpan = 3
```
smeijer, line 94: When the user is logged in, the team ring exists of 4 segments. 2 for the user their team, and 2 for the others. This variable controls the relative size of the user's segments. A 3 makes the "current team" 3 times as large as the other teams.

image

app/components/hooks/use-element-state.tsx
```diff
@@ -0,0 +1,86 @@
+import {useRef, useEffect, useState, RefCallback, useCallback} from 'react'
+
+type State = 'active' | 'focus' | 'hover' | 'initial'
+
+// This started as a work around for https://github.com/framer/motion/issues/1221,
+// but it's so much more now. The variants in framer motion support hover, focus
+// and tap, while this effect also listens to the keypress, so that `Enter`
+// results in an active state as well.
```
smeijer, line 8: This hook will also help with further improvement of the focus state of arrow links/buttons.

kentcdodds (approved):
I love this! Thanks for working on it!

PR #97, opened 2021-07-27
feat: add 'help link' to team stats

smeijer:
Link still needs to point to some page.

Alternative ideas, instead of the text we could render a [`(?)`](https://iconic.app/help-circle/) or [`(i)`](https://iconic.app/information/) svg. Maybe a bit more subtle.

https://user-images.githubusercontent.com/1196524/127175550-a8e78b82-636e-4a75-bb75-874e78bec634.mp4

### DIFF
--- a/app/components/team-stats.tsx
+++ b/app/components/team-stats.tsx
@@ -127,11 +127,27 @@ function TeamStats({
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
This is perfect 👍

I'll get a page and update the href here. Thanks!

PR #98, opened 2021-07-27
feat: add component to show notification messages

smeijer:
Added a component that can be used to show notification messages. Note, that I have not added it to any pages! 

It's based on [`QueryParamNotificationMessage`](https://github.com/kentcdodds/kentcdodds.com/blob/f0e79f337096d0cd695643d292b5e425ca7d0c9d/src/components/notification-message.js#L90-L116) from the current site, with a few changes:

- the message can be rendered in two positions `top-center` and `bottom-right` (default).
- the message can get `autoClose={false}` as argument, so that the user must close it explicitly
- messages are shown one second after page load, to catch that extra bit of attention

My thoughts, render app notifications like `subscribed | form saved` etc `bottom-right` with default settings. And promotional messages like `epic discount!` in `top-center` with `autoClose={false}`.

### DIFF
--- a/app/components/notification-message.tsx
+++ b/app/components/notification-message.tsx
@@ -1,0 +1,78 @@
+import * as React from 'react'
+import {useState} from 'react'
+import {AnimatePresence, motion} from 'framer-motion'
+import clsx from 'clsx'
+import {useRequestInfo} from '../utils/providers'
+import {PlusIcon} from './icons/plus-icon'
+
+function NotificationMessage({
+  queryStringKey,
+  visibleMs = 8000,
+  autoClose,
+  children,
+  position = 'bottom-right',
+}: {
+  queryStringKey?: string
+  children?: React.ReactNode | React.ReactNode[]
+  position?: 'bottom-right' | 'top-center'
+} & (
+  | {autoClose: false; visibleMs?: never}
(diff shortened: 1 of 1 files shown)


app/components/notification-message.tsx
```diff
@@ -0,0 +1,78 @@
...
+              'bottom-8 justify-end': position === 'bottom-right',
+              'top-8 justify-center': position === 'top-center',
+            },
+          )}
+        >
+          <div className="bg-inverse text-inverse relative p-8 pr-14 max-w-xl rounded-lg shadow-md pointer-events-auto">
+            <button
+              aria-label="remove message"
+              onClick={() => setIsVisible(false)}
+              className="text-secondary hover:text-inverse focus:text-inverse absolute right-4 top-8 transform rotate-45"
```
kentcdodds, line 66: Haha, clever way to turn the plus icon into a close icon 😆

kentcdodds (approved):
This is super 👍 Thank you!

app/components/notification-message.tsx
smeijer: Yeah, I didn't feel for adding yet another icon while having something so close :sweat_smile:.

PR #105, opened 2021-07-29
feat: add focus style to sort chat/call buttons

smeijer:
Fixed the hover and focus style of the sort buttons on the `/chat` and `/call` pages, by adding a background color.


kentcdodds (approved):
Good choice I think 👍

PR #106, opened 2021-07-29
feat: remove tap highlight for touch devices

smeijer:
On android, the browser shows a light-blueish overlay on the tapped element. As all our elements do something on focus or tap, we already have something better. This removes the native style, a lot like removing the native focus outline.


kentcdodds (approved):
Yes!

PR #108, opened 2021-07-29
feat: improve focus styles call recorder

smeijer:
- use a background color instead of outline on accordion
- use the hover styles instead of outline on the `delete` button


kentcdodds (approved):
Nice polish!

PR #109, opened 2021-07-29
feat: add motion to the hero

smeijer:
I took the liberty to add some animation to the hero. Please feel free to close it if you don't like what you're seeing. 

It's a bit slow on when remix is building the cache, but I assume that the user will practically never be in that state.

### DIFF
--- a/app/components/notification-message.tsx
+++ b/app/components/notification-message.tsx
@@ -8,6 +8,7 @@ import {PlusIcon} from './icons/plus-icon'
 function NotificationMessage({
   queryStringKey,
   visibleMs = 8000,
+  visible,
   autoClose,
   children,
   position = 'bottom-right',
@@ -15,12 +16,14 @@ function NotificationMessage({
   queryStringKey?: string
   children?: React.ReactNode | React.ReactNode[]
   position?: 'bottom-right' | 'top-center'
+  // make the visibility controlled
+  visible?: boolean
 } & (
   | {autoClose: false; visibleMs?: never}
   | {visibleMs?: number; autoClose?: never}
 )) {
-  // show the message a second after the page has loaded
-  const delay = 1
+  // how long to wait before the message is shown, after mount
+  const delay = typeof visible === 'undefined' ? 1 : 0
   const {searchParams} = useRequestInfo()
   const params = new URLSearchParams(searchParams)
   const [isVisible, setIsVisible] = useState(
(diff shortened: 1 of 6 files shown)


smeijer:
It also removes the pending state from `<body>`, so maybe don't merge this before there is a solution to that.

kentcdodds:
For the pending state, maybe we could just add `cursor-wait` on the `body` (but after a 400ms delay). 

Oh, but that actually wouldn't work for mobile... Hmmm.... https://docs.remix.run just uses a spinner icon in the top right corner... Maybe that would work? Hopefully once we get to start optimizing some things people won't see it very often anyway.

smeijer:
> https://docs.remix.run just uses a spinner icon in the top right corner... Maybe that would work?

Oooh, that gives me an idea. How about we fade the page to the background color, and show the team circle (the one around the profile image) at the exact spot of where it always is, but without profile image, and turning clockwise.

kentcdodds:
> How about we fade the page to the background color

You mean completely hide what the user's looking at? I don't think I want to do that. Maybe fade the page slightly (like we do now), but I actually really like that remix doesn't transition you immediately because what's on the page before the transition can be quite handy while you wait.

I like the idea of reusing the team circle. We just need to keep in mind that the user might not be transitioning to a new page (maybe they just updated their profile info) so we can't assume they'll want to be scrolled to the top of the page. So I'm not sure where we'll have it show up, but I think keeping that theme of the team circle is a good idea.

We'll also want to make sure there's a 300-400ms delay before we show any pending state.

smeijer:
That makes sense. I'll give it another thought.

smeijer:
How about using the notification message for it?

https://user-images.githubusercontent.com/1196524/127638646-47e0c061-0515-42bc-aeb6-dd9478979c14.mp4

smeijer:
A longer video with no delay and extra-long persistence time to show the effect better.

https://user-images.githubusercontent.com/1196524/127639076-5df0036f-9d87-46eb-8499-d9696d682bfd.mp4

kentcdodds:
I love it 🤩 that animation on the transition is fantastic as well!

kentcdodds (approved):
This is sweet! You're creativity has been so great!

kentcdodds:
Ah, merge conflict

smeijer:
> Ah, merge conflict

resolved

kentcdodds:
Super!

PR #110, opened 2021-07-29
feat: improve appearance of twitter embeds

smeijer:
Worked with the examples I had in `the-testing-trophy-and-testing-classifications`. Let me know if I've skipped some cases.


kentcdodds:
Sweet! It's looking way better than what I had 😅

Just a few bits of feedback:

image

Probably want to have the username drop to below the name. And there's no focus/hover styles.

image

Quoted tweet profiles should look the same (just smaller) as regular tweets.

image

I think the margin is off for quoted tweets.

image


Should we worry about the 4 layers of borders we have here? Maybe some padding or something? I think the corners of nested borders in a quoted tweet has a funny look as well.

image

Let's change this to be the same order it is on twitter.com: reply, retweet, like.

And actually, it may be nice to have a fourth icon there that just takes you to the tweet itself. Otherwise it may be non-obvious to people how you can get to the tweet without liking/replying/retweeting (even though you can click any of those + the timestamp to get to the tweet).

image

Love the circle 👍 Let's do a subtle opacity animation on focus/hover.

image

Let's format away the seconds on that time. Just hour and minute is enough.

kentcdodds:
Oh, and this is the page that I was on for my testing: http://localhost:3000/blog/the-testing-trophy-and-testing-classifications

smeijer:
That author name is a bit weird, a cache issue?

https://github.com/kentcdodds/remix-kentcdodds/blob/9e783306769b7cbe67023bb20d96c7a8669118f8/styles/prose.css#L270-L274
image

kentcdodds:
Yup, my bad. Cache issue 😅

smeijer:
So, from all the feedback, only the border issue is valid? :sweat_smile:

PR #111, opened 2021-08-01
feat: keep elements within max content width

smeijer:
- limit max width of the navbar, to be just slightly wider than the content
- adjust max right position of notifications, to be aligned with the right side of navbar
- limit max width of the background of featured sections, to match navbar width

small fixes:
- improve aria-label on "dismiss notification button"
- change size of hero image on login screen


kentcdodds (approved):

kentcdodds:
Super

PR #119, opened 2021-09-22
Fix testing javascript course link

andrewli-ca:
Fixes the external linking to the Testing JavaScript course.


kentcdodds (approved):
Good catch! Thanks!

Issue #121, opened 2021-09-22
Footer at 1024px


Aprillion:
The footer columns look sub-optimal around 1024px window width (Firefox 92, MacOS):

<img width="1060" alt="Screenshot 2021-09-22 at 15 54 03" src="https://user-images.githubusercontent.com/1087670/134357287-01cfccc1-95ad-440e-a695-9147d7fde363.png">

Especially the padding for the email field looks odd - it still seems to be functional, so just a cosmetic issue.

kentcdodds:
How'd your "Learn more" turn green? 🤔 It should be your team color! Weird!

Yeah, I'll give this a look. Maybe we'll just have the newsletter thing take up a full row above the rest of this stuff.

smeijer:
It's might be good enough to just switch to the "tablet breakpoint" sooner. 

![Screenshot_20210922-182339](https://user-images.githubusercontent.com/1196524/134383527-582a3a76-e0f4-4aeb-8a15-dc0ebcdeb193.jpg)

Aprillion:
ah sorry about the color, I was messing up with CSS variables investigating #122 (the team color works fine on all pages except he blog)

kentcdodds:
> It's might be good enough to just switch to the "tablet breakpoint" sooner.

Do you know how to do that @smeijer? 😅

smeijer:
Would you believe me if I said no? :sweat_smile: 

Pull at #132

kentcdodds:
Fixed. Thanks!

PR #132, opened 2021-09-22
fix: change footer breakpoint for 1024px display

smeijer:
This is a quick fix for #121.

I've added an extra breakpoint `xl`, and replaced the `lg:` modifiers in the footer with `xl:`. I've also replaced the `Grid` with a local div, because grid uses the `lg:` to determine the number of columns. (`md` has 8, while `lg` has 12). So that needed to be corrected in the footer.

Meaning, the footer now has a different grid system than the rest of the site. Which I don't really like, but no one might ever notice.

Diff is best viewed with whitespace disabled.

---

Some background; the site designs are optimized for a minimum resolution of 1500px wide. Which is quite a screen. That would mean that most of us would have to experience the site in tablet mode. Hence the current breakpoints. I think that if we need to add more optimizations (likely), it might be a good thing to introduce a breakpoint between `md` and `lg`. If you plan to do so, here's what needs to be done:

1) search + replace `lg:` with `xl:` through all files
2) use the now unused `lg:` modifier to optimize things between tablet (`md`) and standard (`xl`)
3) optionally, tune the pixels in tailwind.config.js

https://github.com/kentcdodds/remix-kentcdodds/blob/4337393e784f9ca70e7bbb58a3b9763dcb29e3c0/tailwind.config.js#L14-L18

### DIFF
--- a/app/components/footer.tsx
+++ b/app/components/footer.tsx
@@ -5,7 +5,6 @@ import {AnchorOrLink} from '~/utils/misc'
 import {useRootData} from '~/utils/use-root-data'
 import {ConvertKitForm} from '../convertkit/form'
 import {H4, H6, Paragraph} from './typography'
-import {Grid} from './grid'
 import {GithubIcon} from './icons/github-icon'
 import {TwitterIcon} from './icons/twitter-icon'
 import {YoutubeIcon} from './icons/youtube-icon'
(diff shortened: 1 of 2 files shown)


kentcdodds:
Thank you! I probably will eventually alter the screen sizes a bit in the future. I appreciate the help here and everywhere :)

PR #134, opened 2021-09-23
Add credits data

dealingwith:
This adds data to the `credits.yml` file. I'm attaching the profile images here for uploading to cloudinary:

Cree
cree

Haze
haze

Kevin
kevin

Maggie
maggie


kentcdodds (approved):
Super, thanks!

PR #141, opened 2021-10-04
feat: make youtube fit screen while maintaining aspect ratio

smeijer:
With this change, the video is no longer cropped when viewing on ultra-wide resolutions.

## Before
https://user-images.githubusercontent.com/1196524/135857438-473d3acf-e56f-44ff-95e7-281c0ddae68e.mp4

## After

https://user-images.githubusercontent.com/1196524/135857292-17e73ea0-3fae-4e40-8b52-1d0af32d79d9.mp4


kentcdodds (approved):
This is way better. Thanks!

PR #142, opened 2021-10-04
Remove duplicated entry in the Shout-outs section inside the credits page

HorusGoul:
I noticed the Fly.io mention was duplicated, this PR removes one of these entries.


HorusGoul:
TIL you can create an empty PR. I think @kentcdodds fixed it at the same time I opened the PR xD That was fast!

kentcdodds:
Thanks anyway! You were quick! :)

PR #147, opened 2021-10-04
Fix typo: existance => existence

Schniz:
Although “exists” or even removing completely the “in existence” could keep the same meaning

great post by the way 👏


kentcdodds (approved):
Thanks!

PR #150, opened 2021-10-05
fix: Center DarkModeToggle icons if browser font size is changed

eps1lon:
Before:
kentcdodds com_toggle previous

After:
kentcdodds com_toggle next

How to reproduce:
Customize font size in the browser (e.g. Chrome > Settings > Appearance > Font Size > Large)


kentcdodds (approved):
Awesome, thanks!

PR #159, opened 2021-10-07
Fix: renamed singIn methods to signIn

danielgary:
Changed:
- [x] Renamed singIn methods to signIn


yarn-error.log
```diff
@@ -0,0 +1,234 @@
+Arguments:
```
kentcdodds: Could you get rid of this please?

kentcdodds (commented):
Nice! Thanks. One thing.

yarn-error.log
danielgary: Already gone

kentcdodds:
Sweet! Thanks!

PR #168, opened 2021-10-10
"Candy dispenser" dark theme fix

NZainchkovskiy:
At ["When to useMemo and useCallback" article](https://kentcdodds.com/blog/usememo-and-usecallback) there is a live demo of "Candy dispenser". I found that div wrapper of that demo has inline styles (especially background color and border) that doesn't respect dark theme.
This quick fix makes it a little bit better.


kentcdodds (approved):
Thanks! That's definitely an improvement 👍

PR #175, opened 2021-10-12
Update testimonials.yml

sergiodxa:
The company I work for changed the name since I wrote the testimonial 😅


kentcdodds (approved):
Coolio 👍

This'll be updated in ~10 seconds :)

PR #177, opened 2021-10-13
fix Peter's twitter handle

Aprillion:
looks like twitter doesn't redirect to updated handles ¯\\\_(ツ)_/¯


kentcdodds (approved):
Did I get it wrong or did you change it?

Aprillion:
I changed it during ~summer. I believe either of us would totally notice if it was wrong at the time of publishing 😅

PR #179, opened 2021-10-13
Update Remix Buy URL

malixsys:
Old URL 404's


smeijer:
Given their recent announcement, I think that paragraph is due to be rewritten.

kentcdodds (approved):
Thanks for the ping. I've updated it. I'm not going to say whether or not that's my actual license key that you could totally use to try things out until it's officially released and open source 🤫

PR #180, opened 2021-10-13
Change "multi-facited" to "multi-faceted"

citizen-xyz:
I am not a native English speaker but so far I've seen the word written only as "multi-faceted".
Also, I've made my research, and Chrome, Google Search, and Grammarly suggest that it should be written as "multi-faceted".
I believe this is the right one and I am also learning something from this! :sweat_smile:


kentcdodds (approved):
Thank you!

PR #183, opened 2021-10-15
Add <link> for items in RSS feed

vaaralav:
Fixes opening blog posts from RSS readers that expect blog posts to have links


kentcdodds (approved):
Oh man! Nice catch. Thank you 👍

PR #185, opened 2021-10-16
Update common-mistakes-with-react-testing-library.mdx

nareshbhatia:
Added note about eslint-plugin-testing-library being available in CRA already.


kentcdodds (approved):
Awesome 👍 Thank you!

PR #186, opened 2021-10-17
fix color of strong link

Aprillion:
Color of `<strong>` was winning over the color of `<a>` in the previous version, hiding the existence of a link:

image


kentcdodds (approved):
Thanks! That's leftover from when I migrated from medium years ago! 😆

PR #187, opened 2021-10-17
fix link to toHaveBeenNthCalledWith

Aprillion:
Jest 22 already had https://archive.jestjs.io/docs/en/22.x/expect#nthcalledwithnthcall-arg1-arg2- 4 years ago, so no excuse for  "unfortunately there is not currently ..."

Though feel free to update the code example too and decline this minimalistic PR  😅


kentcdodds (approved):
Super, thank you for this!

PR #197, opened 2021-10-24
Fix punctuation in CourseCard description: pro's 👉 pros

cmoel:
This is a copy edit change to remove an unneeded apostrophe.


kentcdodds (approved):
Thank you 👏

PR #199, opened 2021-10-25
Removed link to retired repo

dburnside22:
Looks like the repo was retried


kentcdodds (approved):
Good catch! I've updated a few things about this. Thank you 👍

PR #202, opened 2021-10-26
Added Chinese Translation to post "How to use React Context effectively"

davidqqq:
Awesome post!  Hope this translation can help more people to use context effectively :)).


kentcdodds (approved):
Thanks!

PR #210, opened 2021-11-02
Suggestion on how to make error not throw an error

benwainwright:
> I'll bet someone who reads this could figure out a way to make that code throw a different kind of error than a regular Error object (PRs to this post with an example are welcome).

Challenge accepted :)


kentcdodds (approved):
Sure! Thanks :)

PR #212, opened 2021-11-03
Add spanish translation

infoxicator:
I [wrote about DRY recently](https://www.infoxicator.com/the-case-against-dry-micro-frontends-edition) and didn't realise it had a name...AHA!  loved that concept so much that thought it was a pity there wasn't an article in Spanish explaining it, so went ahead and translated it 😁


kentcdodds (approved):
Super! 👏 Thanks :)

PR #219, opened 2021-11-13
Change `the default` to `by default`

sairajchouhan:


kentcdodds (commented):
I think it makes sense as-is and I prefer it that way. Thanks though!

PR #235, opened 2021-11-22
Replace the Codesandbox

SamiKamal:
Replace the *broken* Codesandbox with a link that leads to the exercise file in Github


kentcdodds (approved):
Thank you!

PR #237, opened 2021-11-25
Return null theme session

abereghici:
The dark theme is returned from `getThemeSession` as fallback. This way the client script that detects the user theme isn't executed at all.

### DIFF
--- a/app/utils/theme.server.ts
+++ b/app/utils/theme.server.ts
@@ -20,7 +20,7 @@ async function getThemeSession(request: Request) {
   return {
     getTheme: () => {
       const themeValue = session.get('theme')
-      return isTheme(themeValue) ? themeValue : Theme.DARK
+      return isTheme(themeValue) ? themeValue : null
     },
     setTheme: (theme: Theme) => session.set('theme', theme),
     commit: () => themeStorage.commitSession(session),


kentcdodds (commented):
I'm not sure I understand why this is desirable 🤔

abereghici:
@kentcdodds In `root.tsx` loader, `themeSession.getTheme()` (https://github.com/kentcdodds/kentcdodds.com/blob/main/app/root.tsx#L188)  will always return a value ( the stored theme or dark theme ). This means the ssrTheme will be true every time here:

```
<NonFlashOfWrongThemeEls ssrTheme={Boolean(data.requestInfo.session.theme)}  />
``` 
and the client code that should detect the current client theme won't be executed. ( https://github.com/kentcdodds/kentcdodds.com/blob/main/app/utils/theme-provider.tsx#L163 ) 


I have a white theme now, and if I open kentcdodds.com in incognito tab, I have the black theme displayed instead of white, because the client side code didn't run to detect the right theme.

kentcdodds:
Right, this is intentional. The default theme for the site is dark so if you haven't set a theme value for your OS, then I want it to be dark. It only will go with light if your system is set up with [`(prefers-color-scheme: light)`](https://github.com/kentcdodds/kentcdodds.com/blob/main/app/utils/theme-provider.tsx#L20). I'm not sure the implications of that resolving to true by default on all systems, but the site looks better in dark mode so we default to dark mode and only if the system explicitly matches that media query or the user explicitly chooses light mode do we use light mode.

abereghici:
@kentcdodds I'm trying to understand when the client code is executed.  If we always return a theme from server session, it means `specifiedTheme` will always contain a value and `getPreferredTheme()` won't be called.

The comment also says `On the server, if we don't have a specified theme then we should return null and the clientThemeCode will set the theme for us`

I thought from session server we return the stored user preference if any, otherwise we let the client code to decide which theme should be displayed ( and the dark theme should be set as default there )

```
function ThemeProvider({
  children,
  specifiedTheme,
}: {
  children: React.ReactNode
  specifiedTheme: Theme | null
}) {
  const [theme, setTheme] = React.useState<Theme | null>(() => {
    // On the server, if we don't have a specified theme then we should
    // return null and the clientThemeCode will set the theme for us
    // before hydration. Then (during hydration), this code will get the same
    // value that clientThemeCode got so hydration is happy.
    if (specifiedTheme) {
      if (themes.includes(specifiedTheme)) return specifiedTheme
      else return null
    }

    // there's no way for us to know what the theme should be in this context
    // the client will have to figure it out before hydration.
    if (typeof window !== 'object') return null

    return getPreferredTheme()
  })
```

I learn Remix now and looking at your codebase it's an awesome way to do it.

kentcdodds:
> the clientThemeCode will set the theme for us

https://github.com/kentcdodds/kentcdodds.com/blob/3b35752fc400f8785ec94aa6a75c6d006b4506cb/app/utils/theme-provider.tsx#L105

> this code will get the same value that clientThemeCode got so hydration is happy.

https://github.com/kentcdodds/kentcdodds.com/blob/3b35752fc400f8785ec94aa6a75c6d006b4506cb/app/utils/theme-provider.tsx#L92

https://github.com/kentcdodds/kentcdodds.com/blob/3b35752fc400f8785ec94aa6a75c6d006b4506cb/app/utils/theme-provider.tsx#L22

That's the basics of how it all works. No flash of incorrect theme :)

PR #239, opened 2021-11-28
If the font family is already installed, use it.

shamwela:
If the "Matter" font family is already installed on the user's device, it will be used instead of downloading the font file.


kentcdodds:
I don't think anyone will have it installed though 🤔 and what if they have a different font by the same name?

shamwela:
I didn't think about those. I agree with you. You can close this now. Thanks.

shamwela:
I didn't think about those. I agree with you. I will close this now. Thanks.

PR #243, opened 2021-12-04
fix: server navigation to `talks/$slug`

dodas:
Fixes server returning `null` when navigating to any `/talks/something` route: https://github.com/kentcdodds/kentcdodds.com/pull/82#issuecomment-985932847

https://user-images.githubusercontent.com/14360171/144715903-fa72194f-dc2f-4f69-bc5d-5fec60d404c3.mov


kentcdodds (approved):
Perfect. Thank you!

PR #246, opened 2021-12-07
Update CONTRIBUTING.md

dane-stevens:
fix: docker-compose command


kentcdodds (commented):
I believe Docker added the `compose` command in a recent version. Maybe update your docker version?

dane-stevens:
Ahh! You're right!

Thank you.

PR #248, opened 2021-12-08
feat: added italian translation for the "TypeScript Function Syntaxes" post

BobToninho:
I'm just starting to create my blog and I thought a good article like this could be good for it, let me know if there is any kind of problem :)


kentcdodds (approved):
Super 👏 Thanks!

PR #250, opened 2021-12-09
fix: call recorder timer

dane-stevens:
- prettier formatting
- updated use-interval to typescript
- blinking timer text at 01:45, fast blink at 02:00, no blink at 02:10


kentcdodds:
Awesome! I'm not sure I want the blinking animation. I don't want to give people anxiety about going a bit long 😅

dane-stevens:
Lol, I had the same thought. I'll remove it then!

kentcdodds (approved):
Stellar. Thanks!

PR #251, opened 2021-12-19
fix: use `-dd` for day of month rather than `-ii` for talks

lukebrowne:
Spotted that the Brum.js date was incorrect so checked the data was correct and spotted it was using the wrong value for day of month.

I've also updated the Brum.js website to be our preferred site rather than Meetup :)

I couldn't see that you're doing tests for these components, so didn't add -- just let me know if you'd rather that

**Before:**
image

**After:**
image


kentcdodds (approved):
Thank you for this 👏

PR #253, opened 2021-12-25
chore: update `deployment.yml`

MichaelDeBoey:
No need to pass the `access_token` anymore, as it defaults to `secrets.GITHUB_TOKEN`


kentcdodds (approved):
Thanks! And Merry Christmas! 👼

PR #258, opened 2021-12-26
chore: update `deployment.yml`

MichaelDeBoey:
Since https://github.com/FedericoCarboni/setup-ffmpeg/pull/6 is merged, we don't need to pass this anymore


kentcdodds (approved):
Thanks!

PR #259, opened 2021-12-26
chore: remove `REMIX_TOKEN` references

MichaelDeBoey:


other/docker-build
```diff
@@ -1 +1 @@
-docker build . --build-arg REMIX_TOKEN=${REMIX_TOKEN}
\ No newline at end of file
+docker build .
```
kentcdodds: Wonder if we can get rid of the docker build script file

kentcdodds (commented):
Do you know if that file was used anywhere?

MichaelDeBoey:
@kentcdodds I checked for code references, but couldn't find any usage

kentcdodds:
Thanks!

MichaelDeBoey:
@kentcdodds Build & deploy seems to work 👍
https://github.com/kentcdodds/kentcdodds.com/runs/4637430239

PR #262, opened 2021-12-28
Change the Counter component by the KeyPropResetCounter component (TS)

jpmti2016:
Change the code to include the components that are used in the KeyPropReset example.

The KeyPropReset component use the KeyPropResetCounter component, not the Counter component.

With TypeScript this time (my bad).


kentcdodds (approved):
Thanks!

PR #263, opened 2022-01-02
Checkout vs Check out

gpbmike:
Checkout is a noun, check out is the verb.

image


kentcdodds (approved):
Yep, that makes sense 👍 thanks!

PR #282, opened 2022-01-23
Updated course link to EggHead

Hillsie:
Kent C. Dodds, 
A quick PR with an updated EggHead course link, while reading.


kentcdodds (approved):
Thanks 😊

PR #283, opened 2022-01-24
chore: use official `prettier-plugin-tailwindcss` instead of `prettier-plugin-tailwind`

MichaelDeBoey:
https://tailwindcss.com/blog/automatic-class-sorting-with-prettier

I also ran Prettier on the full codebase


kentcdodds (approved):
I'm definitely a fan! Thank you!

PR #285, opened 2022-01-27
fix: changing brazilian portuguese translation to use-effect-use-layo…

allangrds:
Changing the label to one already used in other blog posts


kentcdodds (approved):
👍

PR #287, opened 2022-02-01
`that take some time` → `that can take some time`

jasikpark:


kentcdodds (commented):
Thanks for this! It reminds me that I improved local development quite a bit and it's not as complicated anymore. I'll go ahead and update this now.

kentcdodds:
Actually, I can still use your change :)

jasikpark:
Oooh nice - I was sleuthing through to look at your site dev flow - I've really been wanting to improve the simplicity for hosting a simple blog that posts photos and text, like a basic WP blog but w/ git-backed content management... Though I then remembered that you decided to use `github.com` as your CMS 😆 

(currently tempted to just spin up a ghost blog)

kentcdodds:
Yeah, the benefit of using GitHub as a CMS is that people can contribute fixes to it without me having to worry about user management etc. It's quite nice :)

PR #290, opened 2022-02-06
Update intentional-career-building.mdx

salkanovicsafet:
Remove dead link


kentcdodds (approved):
Thanks! I can't even remember what that was about 😅 shame it got deleted

PR #293, opened 2022-02-13
add pt-BR translated version

biantris:
Adding the translation of the great article React Fundamentals: Props vs State to pt-BR so we can spread more good knowledge in the Brazilian community \o/

- [x] Added title and links to the original article and Kent C. Dodds website as requested.


kentcdodds (approved):
Sweet!

PR #295, opened 2022-02-15
Add spanish translation

sebagnz:
Add spanish translation for "Fix the slow render before you fix the re-render" post.


kentcdodds (approved):
Awesome! Thanks :)

PR #298, opened 2022-02-24
Fix result scope

velusgautam:
This PR makes a small edit which changes the scope of `result` variable.


kentcdodds (approved):
Sure 👍 Thanks :)

PR #300, opened 2022-03-05
Add scrollbar color

DhrumanGupta:
I noticed that the website has the default scrollbar. It looks a little ugly (especially since the website looks so amazing!), so I added some colours that align with the theme there. 

Note: this only shows on webkit browsers

### DIFF
--- a/styles/tailwind.css
+++ b/styles/tailwind.css
@@ -70,3 +70,18 @@
     @apply text-white dark:text-black;
   }
 }
+
+body::-webkit-scrollbar {
+  @apply w-2;
+}
+body::-webkit-scrollbar-track {
+  @apply rounded bg-white shadow-none dark:bg-gray-900;
+}
+
+body::-webkit-scrollbar-thumb {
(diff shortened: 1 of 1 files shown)


DhrumanGupta:
I've added some screenshots too!  
Screenshot_20220306_002544
Screenshot_20220306_002533

kentcdodds:
Thanks for this! Could you provide a screenshot of what it looks like before these changes? I'm on a Mac so the scroll bar isn't visible unless I'm scrolling and it looks nice to me. What does it look like for you?

DhrumanGupta:
Yeah sure!  
Screenshot (1)
Screenshot (2)
  
I've made the sceollbar thinner and removed it's background so it blends well

kentcdodds:
Thank you for putting the time in for this. However I don't like the way this shifts the content of the page over to the left by the width of the scroll bar and I don't like that it looks worse on my own computer (a Mac). So I don't think I'll merge this change. Thanks anyway!

DhrumanGupta:
Oh yeah sure! Makes sense. Thanks for taking the effort to try everything out!

PR #306, opened 2022-03-16
add Chinese (simplified) translation of Colocation

SevenOutman:
Hi Kent, thanks for this great post! I've translated it into Chinese and [posted it on juejin.im](https://juejin.cn/post/7075233317284479013), a popular content-based developer community (name in Chinese meaning Gold Mining) in China so that more people can read it.


kentcdodds (approved):
Thanks!

PR #307, opened 2022-03-17
fix isDeployable check

SevenOutman:
Hi Kent, it looks like some changes in your `content/` directory were not recognized as deployable in your GitHub Actions ([see logs](https://github.com/kentcdodds/kentcdodds.com/runs/5574387062?check_suite_focus=true#step:4:24)). I looked into the `is-deployable.js` script and noticed this possible mistake.

### DIFF
--- a/other/is-deployable.js
+++ b/other/is-deployable.js
@@ -11,14 +11,15 @@ async function go() {
     compareCommitSha,
     changedFiles,
   })
-  // deploy if:
+  // don't deploy if:
   // - there was an error getting the changed files (null)
   // - there are no changed files
-  // - there are changed files, but at least one of them is non-content
-  const isDeployable =
+  // - there are changed files, but all of them are non-content
+  const isDeployable = !(
     changedFiles === null ||
     changedFiles.length === 0 ||
-    changedFiles.some(({filename}) => !filename.startsWith('content'))
+    changedFiles.every(({filename}) => !filename.startsWith('content'))
+  )
 
   console.error(
     isDeployable


kentcdodds:
Hi! You're looking at the wrong GitHub action. That one deploys a new version of the app. We don't want to deploy a new version of the app when just the content changed. We have a different action for that: https://github.com/kentcdodds/kentcdodds.com/runs/5574387059?check_suite_focus=true

Thanks anyway :)

SevenOutman:
Hi Kent, there's still "no translations" shown [on the post page](https://kentcdodds.com/blog/colocation), even after I clear the cache. Did I miss something?

kentcdodds:
Hmmm... I'm not sure what the problem could have been, but now I see it:

image

And based on this it did get an update: https://github.com/kentcdodds/kentcdodds.com/runs/5574387059?check_suite_focus=true#step:4:22

kentcdodds:
Looking at that github action further it's showing a couple other pages that you did not touch, indicating that they haven't been updated until this time which is unusual. I'm thinking there is some broken logic in my refresh code that's causing this problem.

SevenOutman:
> Hmmm... I'm not sure what the problem could have been, but now I see it:

Looks like some CDN problem?

kentcdodds:
I figured out the problem: https://github.com/kentcdodds/kentcdodds.com/commit/de0a0766374f909469950649b79fdaf2938c8e46

I was JSON.stringifying this data and then that was JSON.stringified again before going into the cache. So when it came out of the cache it was JSON.parsed only once and the result was a string instead of an object. This should not be a problem going forward. Thanks!

PR #309, opened 2022-03-24
feat: update appearances

MichaelDeBoey:
Don't know if they're in the right category, but they're not really tests or podcast so figured the `Interviews & Chats` section was best 🤔


kentcdodds (approved):
Somehow totally forgot to add this one. Thanks!

PR #321, opened 2022-04-24
add 'the-state-reducer-pattern' Korean translation

woobottle:
Adding Korean translation for a posts:

* the-state-reducer-pattern


kentcdodds (approved):
Thanks!

PR #323, opened 2022-04-25
Add Translate 'Write fewer, longer tests'

danywalls:
Add 'write-fewer-longer-tests' Spanish translation


kentcdodds (approved):
Awesome! Thanks 😊

PR #324, opened 2022-04-25
edit typo in fix-the-slow-render-before-you-fix-the-re-render

woobottle:
image

i found jsx without code block
if it was your intend it need no fix
but i think this is bug :)
so i make pr for fix this typo


kentcdodds (changes requested):
This is intentional. That bit of the post actually renders something interactive people can click on. This is utilizing the power of MDX.

Thanks anyway!

PR #326, opened 2022-04-25
Fix the frontmatter structure and add extra information.

danywalls:
Fix the frontmatter structure and add extra information.

> After seeing it https://kentcdodds.com/blog/write-fewer-longer-tests, the link doesn't show as expected.


kentcdodds (commented):
👍👍

PR #329, opened 2022-04-28
Syntax that works with return values

eric-burel:
Hi,
It seems that after updating to React 17, the `act` function accepts only a "void" return. However, if your promise must return a value, for instance a mock of a success message from the server, the current code will throw a TypeScript error, because the promise have a type different from undefined/void.
This version is safer (yet a bit more verbose, I personally don't get this choice in React...).

You might want to double check though, I've noticed that in the context of a massive migration, and I didn't test React 18 either


kentcdodds (approved):
Thanks 👍

PR #331, opened 2022-05-06
Translation of get-a-catch-block-error-message-with-typescript to pt-BR

Ruttmann:
This article was so helpful to me that I decided to make a translation to my native language. :)


kentcdodds (approved):
Sweet!

PR #337, opened 2022-05-21
fixing a typo

ricaguzam:
Just happened to see this typo and sent a proposal to fix it


kentcdodds (commented):
Muck is the intended word here actually. Thanks though!

PR #343, opened 2022-06-16
Use Octokit methods for improved typings

Dueen:
Octokit provides methods for downloading files by path and by SHA with associated types for the return value, this makes it easier to check and assert.

### DIFF
--- a/app/utils/blog.server.ts
+++ b/app/utils/blog.server.ts
@@ -143,7 +143,7 @@ async function getMostPopularPostSlugs({
 }) {
   const postsSortedByMostPopular = await cachified({
     key: `sorted-most-popular-post-slugs`,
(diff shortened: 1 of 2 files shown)


w3bdesign (approved):

app/utils/blog.server.ts
```diff
@@ -143,7 +143,7 @@ async function getMostPopularPostSlugs({
 }) {
   const postsSortedByMostPopular = await cachified({
     key: `sorted-most-popular-post-slugs`,
-    ttl: 1000 * 60,
```
kentcdodds, line 146: Whoopsies, didn't mean to include that commit in here 😅 Oh well.

kentcdodds (approved):
Thanks a bunch! This is nicer :)

PR #344, opened 2022-06-20
Update replace-axios-with-a-simple-custom-fetch-wrapper.mdx

dimaMachina:
Simplify, `Promise.reject` can be replaced with `throw`


kentcdodds:
Thanks! I like the way it's written better. But what you have would also work :)

PR #350, opened 2022-07-21
Scrollbar now follows theme when it's is dark

clicktodev:
closes #349
Before:
Screenshot from 2022-07-21 15-35-30
After:
Screenshot from 2022-07-21 15-36-18

PR #351, opened 2022-07-21
Scrollbar color now follows theme (fix)

clicktodev:
This is to fix a regression caused by #350 in which the scrollbar is stuck in dark mode even after changing to light mode.
No idea atm why this happened, It was working fine locally.

### DIFF
--- a/styles/app.css
+++ b/styles/app.css
@@ -36,6 +36,7 @@
   --border-secondary: var(--color-gray-200);
   --text-primary: var(--color-black);
   --text-secondary: var(--color-gray-500);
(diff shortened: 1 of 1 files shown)


clicktodev:
Just tested again and apparently there is different behavior between chrome for Ubuntu and chrome for Windows.
Chrome for Windows requires I explicitly set the `color-scheme: light;` for it to revert back to light.

kentcdodds (approved):
Thanks!

clicktodev:
Anytime! 👏

PR #355, opened 2022-07-29
Update but-really-what-is-a-javascript-mock.mdx

re4388:
I guess this is a typo? muck -> mock


kentcdodds (commented):
Thanks! But no. The intended word here is "muck"

PR #358, opened 2022-08-07
Fix duplicate router issue

andrew-d-moore:
TL;DR: Provide router instance to route methods.

Great article sir. I have one small change that cropped up when I was implementing this solution. You actually declare two routers one in the getRoutes() method and one in the getMathRoutes() method. This was causing an issue where the routes weren't mounting to the server instance because one router was being returned while two were created. So I simply made the change to provide the router instance to the route methods I wanted mounted, I saw you had a github link for it so figured I'd just do a quick PR. 

Keep up the great work.

### DIFF
--- a/content/blog/how-i-structure-express-apps.mdx
+++ b/content/blog/how-i-structure-express-apps.mdx
@@ -338,7 +338,7 @@ function getRoutes() {
   // create a router for all the routes of our app
   const router = express.Router()
 
-  router.use('/math', getMathRoutes())
+  router.use('/math', getMathRoutes(router))
   // any additional routes would go here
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
I don't think this will do what you think it will. I believe it will have a `/math/add` route, but it'll also create a `/add` route as well 😬

andrew-d-moore:
I did some further testing and you're right, it does create a direct route to math as well. I also figured out why I was having the initial mounting issues. If you wanted all direct routes then my solution would work, but that clearly isn't the intention so I'll close this. Sorry about that.

PR #360, opened 2022-08-07
Fix tracking visits in client-side navigation

silvenon:
It took me a while to figure this one out in my own site, so I wanted to share it with you in case you wanted it. 🙂


kentcdodds (commented):
Hmm... Should we instead make it so the cononical URLs change during client-side navigation?

silvenon:
I tried that too using `useMatches`, but Fathom wasn't getting it either, it still tracked visits only on full document requests. Probably because the history change happens right away, while the content (and therefore the canonical) takes some time to load, so Fathom still sees the old one.

kentcdodds:
So does that mean that Fathom won't track client-side navigations?

silvenon:
Actually, what we can do is change canonicals during client side navigation, like you said, and inside a `useEffect` call Fathom's `trackPageview`, passing the canonical as a dependency to the `useEffect`. In order to prevent a double visit on initial page load we'd need to set `auto` to false when loading Fathom.

Also, I'm not sure what is gained by using the `fathom-client` as opposed to just adding the `<script>` tag, which gets rid of those nasty hydration errors in the console.

I edited the code to reflect my new suggestion. I tried this on my own site and it worked for me.

app/root.tsx
```diff
@@ -307,6 +307,26 @@ function PageLoadingMessage() {
...
+declare global {
+  const fathom: {
+    trackPageview(): void
+  }
+}
+
+function CanonicalLink({origin, path}: {origin: string; path: string}) {
+  const matches = useMatches()
+  const leafRoute = matches[matches.length - 1]
+  const canonicalUrl = removeTrailingSlash(
```
kentcdodds: Should this not just be `useLocation().pathname`?

kentcdodds (commented):
This is great! Just one question.

app/root.tsx
silvenon: Actually yes, in my case I was using `useMatches` because I wanted to leave the option to change the canonical URL per route via a loader or handle, but if you don't need this then `useLocation` is fine. Fixed.

app/root.tsx
```diff
@@ -307,6 +307,23 @@ function PageLoadingMessage() {
...
+    trackPageview(): void
+  }
+}
+
+function CanonicalLink({origin}: {origin: string}) {
+  const {pathname} = useLocation()
+  const canonicalUrl = removeTrailingSlash(`${origin}${pathname}`)
+
+  React.useEffect(() => {
+    fathom.trackPageview()
```
kentcdodds: Is it possible this could run before fathom is finished loading? Should we update the types/code to reflect this possibility?

silvenon: I don't know, that's a good question… Maybe it would be better to sacrifice the console and use `fathom-client` instead, since it's specifically designed to be able to track before the script has finished loading.

I updated the code.

kentcdodds:
But doing this will give us the hydration warning again right? I think I'd rather go with what we had before and figure out how to make sure we run things after it's finished loading.

silvenon:
Ok, I copied the technique used in `fathom-client` to queue the command in case Fathom is not loaded yet. I don't know if it works, though, so maybe it's best if you take over from here. Perhaps test it by mocking `https://sailfish.kentcdodds.com/script.js` to respond late in order to test the queue, or something like that.

kentcdodds (approved):
This looks great to me. Thank you!

silvenon:
You're welcome ❤️

PR #367, opened 2022-08-25
Remove // for convenience, but totally not required...

Koboki:
I removed the comment that says "for convenience, but totally not required..." as these three lines of code are actually required and without them the code will throw an error. I'm not sure why Kent wrote that comment. Am I missing something?


Koboki:
I removed the comment that says "for convenience, but totally not required..." as these three lines of code are actually required and without them the code will throw an error. I'm not sure why Kent wrote that comment. Am I missing something?

kentcdodds (approved):
Been wanting to do this for a while

PR #369, opened 2022-09-07
adds Turkish translation

yilmazcite:
I translated the "Remix: the Yang to React's Yin" blog (https://kentcdodds.com/blog/remix-the-yang-to-react-s-yin) into Turkish and referred to the main source at the top of the translation as requested.


kentcdodds (approved):
Thanks!

PR #373, opened 2022-09-22
feat: add Japanese translation for article "Testing Implementation Details"

pandanoir:
Hi, Kent.

I've read Testing Implementation Details and been impressed. I shared it with our team. However, our team consists of Japanese people. So I've translated it into Japanese.


kentcdodds (approved):
Sweet! Thanks 👍

PR #374, opened 2022-09-28
Fix typo in Rest/Spread code snippet

jcat4:
In the code snippet for Rest/Spread in the "JS to know for React" page, each key in the object `obj2` is set to a value of `'{key} from obj2'`.  The `h` key is erroneously being set to `'g from obj2'`.


kentcdodds (approved):
Thanks!

PR #376, opened 2022-10-08
fix syntax error in usememo-and-usecallback

Aprillion:
reported on https://discord.com/channels/715220730605731931/715220730605731934/1028264628280627281


kentcdodds (approved):
Thanks!

PR #378, opened 2022-10-19
fix for incorrect lines highlighted

Svish:
These two highlighted lines should (I'm assuming) be the lines mentioning `cross-env`, above them.

image


kentcdodds (approved):
Thanks!

PR #380, opened 2022-10-23
feat: add japanese translation for aha testing

makotot:
Added Japanese translation for https://kentcdodds.com/blog/aha-testing.


kentcdodds (approved):
Thanks! 👏

PR #382, opened 2022-10-27
feat: add spanish translation for blog post colocation

acontreras89:
Add spanish translation for [Colocation](https://kentcdodds.com/blog/colocation).

The translation can be seen [here](https://redradix.com/blog/Colocalizacion/).


kentcdodds (approved):
Thanks!

PR #391, opened 2022-11-25
Refactor: show the remaining time of black friday

vab1997:
I created component for show different type time in promotion of Black Friday. I hope you like 🙂. 
image


kentcdodds:
Thanks! I prefer to avoid those kinds of abstractions though. Cheers!

PR #403, opened 2023-01-14
feat: add japanese translation for Colocation

makotot:
Added Japanese translation for https://kentcdodds.com/blog/colocation.


kentcdodds (approved):
Thanks!

PR #404, opened 2023-01-16
feat: add japanese translation for aha testing

makotot:
The link added in [this PR](https://github.com/kentcdodds/kentcdodds.com/pull/380) has disappeared, so let me add it again.
It has probably been removed in this [commit](https://github.com/kentcdodds/kentcdodds.com/commit/5f19e7b78912c5328bd180fcc2b7c5ff8f4b7ced), but if it was not intentionally deleted, I would like to have it added again 🙏


kentcdodds (approved):
Thanks!

PR #412, opened 2023-01-28
add kr translate "useEffect vs useLayoutEffect"

junghyeonsu:
hello!
I added **korean** translate "useEffect vs useLayoutEffect"


kentcdodds (approved):
Terrific! Thanks 👍

PR #415, opened 2023-02-02
add kr translate "Get a catch block error message with TypeScript"

junghyeonsu:
hello!
you can check my blog [here](https://junghyeonsu.com/posts/get-a-catch-block-error-message-with-typescript/)
thank you!


kentcdodds (approved):
Great! Thank you :)

PR #417, opened 2023-02-07
add ja translate

kazu-2020:
hello!
I added Japanese translate "prop drilling " and "When to break up a component into multiple components
" 👍🏻

Thanks for the great article!


kentcdodds (approved):
Awesome. Thanks!

PR #425, opened 2023-02-27
added a Japanese translation for Get a catch block error message with TypeScript

Lada496:
Added a Japanese translation for [Get a catch block error message with TypeScript](https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript)


kentcdodds (approved):
Great! Thank you :)

PR #427, opened 2023-03-04
added a Japanese translation for Why I Never Use Shallow Rendering

y-takebe:
I have prepared a Japanese translation of the article.

Thank you for your always informative articles.


kentcdodds (approved):
Thanks!

PR #428, opened 2023-03-05
Add kr translation of "Static vs Unit vs Integration vs E2E Testing for Frontend Apps"

Soojae:
Added korean translation.
Thank you for your awesome post!


kentcdodds (approved):
Awesome. Thank you!

PR #430, opened 2023-03-07
Add kr translation for "How to know what to test"

Soojae:
Added korean translation.
Thanks for sharing such an informative post!


kentcdodds (approved):
Thanks!

Could you update the link back to my post? It looks like you're liking to a specific heading rather than to the top of the page

Soojae:
Updated the link, thanks!

PR #434, opened 2023-03-13
add kr translation for "Testing Implementation Details"

Soojae:
Added korean translation.
Thank you for posting a great article!


kentcdodds (approved):
Thank you!

PR #440, opened 2023-04-08
Update CONTRIBUTING.md

quemazote:
The better-SQLite3 can use many binaries. On some installs it uses the x86_64 binary. If you are on Apple Silicon this becomes a problem. Even though the arm64 is available it doesn't recognize that it should use the arm64. By Uninstalling and installing with the command I added to this document you can resolve that issue. Hopefully the newer versions will address this.


kentcdodds (approved):
Thanks :)

PR #441, opened 2023-04-08
Update testing-implementation-details.mdx

evinracher:
I propose this change because the functionality to have multiple accordion items to be open at once is not working as I think it should. I am not sure if the line: tsx remove=2-3,16,25 add=4-14,17,26 if right, as it if diffucult to check without have the visual help. I hope this make the content more clear. Good post by the way. Thanks for sharing.

P.S. this is the first time I propose change to this project.


kentcdodds:
In the blog post the point is not to implement the multiple at once feature, but to refactor so it's easier to add the multiple at once feature later.

Thanks anyway!

PR #442, opened 2023-04-16
add kr transtion for test-isolation-with-react

J-Ymini:
Added korean translation for test-isolation-with-react.
Thank you for your great article!


kentcdodds (approved):
Thanks!

PR #443, opened 2023-04-22
fix: make sure X is displayed correctly on burger menu

### LINKED ISSUE #43: the close button for mobile menu transforms weirdly
The close button works fine, as long as the page was loaded with the button visible (on narrow screen). But when loading the page in "large screen", then resizing to "narrow screen", and then clicking the button, the transition is malformed.

It feels like a bug in framer-motion, as I can also not force the `transform-origin` on those svg elements.

I expect that this can be fixed by not rendering the button in a hidden state. Will try later.

smeijer:
Took me long enough 😅, but this should fix #43

### DIFF
--- a/app/components/navbar.tsx
+++ b/app/components/navbar.tsx
@@ -171,8 +171,8 @@ function MobileMenuList() {
 }
 
 const topVariants = {
-  open: {rotate: 45, y: 7},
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Oh sweet! Thank you 👏👏

PR #445, opened 2023-04-28
update chinese translation link

finorix079:
The original link is no longer working. It seems the author has rephrased his articles' links on his site.


kentcdodds (approved):
Thank you!

PR #458, opened 2023-06-26
update to native ESM

onemen:
@kentcdodds,

- there are some comment in code review

- need see if all github action are working


mocks/generate/tweets.ts
```diff
@@ -0,0 +1,39 @@
+import 'dotenv/config'
+import path from 'path'
+import {fileURLToPath} from 'url'
+import fs from 'fs'
+import {installGlobals} from '@remix-run/node'
+// NOTE: run this with tsx ./mocks/generate/tweets
+import {getTweet} from '../../app/utils/twitter/get-tweet.ts'
```
onemen, line 7: the old code import `twitter.server` that was not exist
`const {getTweet} = require('../../app/utils/twitter.server')`

onemen (commented):
I did not change any of the mdx files to ESM
for example form `how-i-structure-express-apps.mdx`
```js
if (process.env.NODE_ENV === 'production') {
  require('./dist')
} else {
  require('nodemon')({script: 'dev.js'})
}
```

```js
const pkg = require('./package.json')

module.exports = {
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          node: pkg.engines.node,
        },
      },
    ],
  ],
}
```

kentcdodds (approved):
Great 👍 thanks!

kentcdodds:
Seriously, this is a massive project and I appreciate you taking it on. Thank you so much!

PR #460, opened 2023-06-30
add pagination for calls section

girishsavlesha:
Hi Kent! 👋 , added pagination for the calls sections. right now all episodes have `Season 1` associated with them,  if we get new seasons this will get updated accordingly.

Let me know if you need some more changes to this. 😄 

ezgif-5-a662310dcb


kentcdodds (approved):
Thank you so much for this! I'm going to merge it into `dev` so I can get it deployed and see how it behaves with production data. Thanks!

kentcdodds:
Hi @girishsavlesha! I'm going to do 50 episode seasons. Would you be interested in handling the create case as well? The season [here](https://github.com/kentcdodds/kentcdodds.com/blob/4981f21d083269401974b8ef887ed4cc9b067aca/app/utils/transistor.server.ts#L99C17-L99C17) should be the latest season. If it's going to be the 51st episode then we want to create a new season and make it the 1st episode of that season. If you're not able to I understand. I can get to it later :)

girishsavlesha:
yeah sure! I can look into this. 😃

PR #465, opened 2023-07-01
Create case for 50 episode/seaons for podcast (#460)

girishsavlesha:
#460

Hey @kentcdodds, I am unsure how I can test this on local? 
you can review the code and let me know if any changes are required.


app/utils/transistor.server.ts
```diff
@@ -84,6 +84,15 @@ async function createEpisode({
   })
   const {upload_url, audio_url, content_type} = authorized.data.attributes
 
+  const totalEpisodes = await fetchTransitor<TransistorEpisodesJson>({
```
kentcdodds: Is there any way for us to avoid calling this API? It's just sending a lot of data 🤔

Maybe we can request the seasons first, then query the latest season. Then we'd know from the start what season and episode we should create.

kentcdodds (commented):
Thanks!

app/utils/transistor.server.ts
girishsavlesha: I am unsure how we can request the season, 
https://developers.transistor.fm/#episodes  doc seemed a bit confusing to me. 
I tried to get latest episode and pulled the season out of it, I think that will give us the latest season

app/utils/transistor.server.ts
```diff
@@ -242,6 +259,19 @@ async function getEpisodes() {
   return episodes
 }
 
+async function getCurrentSeason() {
```
girishsavlesha, line 259: @kentcdodds

app/utils/transistor.server.ts
```diff
@@ -96,7 +101,7 @@ async function createEpisode({
       // IDEA: set the season automatically based on the year
       // new Date().getFullYear() - 2020
       // need to support multiple seasons in the UI first though.
```
kentcdodds:
```suggestion
```

kentcdodds (approved):
Looks awesome. Thank you!

PR #467, opened 2023-07-11
add korean translation to stop-being-a-junior post

moonheekim0118:
Here it is a korean translation of your post 'stop being a junior'
thank you for your great contents always


kentcdodds (approved):
Thank you!

PR #468, opened 2023-07-13
fix: broken meta on single call's  episode page

amir-ziaei:
Noticed the meta func is broken on single episodes of Call Kent Podcast.
E.g. https://kentcdodds.com/calls/04/07/remix-full-stack-components-vs-server-components has a page title of "Call not found"
image

Please check this on your side as well because I couldn't fully verify the solution as some e2e tests were already broken.

Thanks!


kentcdodds (approved):
Ah yeah! Nice catch. Thank you :)

PR #472, opened 2023-08-01
chore: install `source-map-support`

MichaelDeBoey:
See https://github.com/remix-run/remix/pull/7009 & https://github.com/remix-run/remix/pull/7010

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "kentcdodds.com",
   "version": "1.0.0",
   "engines": {
-    "node": "18",
+    "node": "16",
     "npm": "8"
   },
   "description": "A remix version of kentcdodds.com",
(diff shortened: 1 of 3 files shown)


kentcdodds:
Thank you!

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "kentcdodds.com",
   "version": "1.0.0",
   "engines": {
-    "node": "18",
+    "node": "16",
```
kentcdodds, line 5: I just noticed this change. Is this intentional?

MichaelDeBoey: Oh no that wasn't intentional, sorry

You can revert that if you want

kentcdodds: No worries :) Thanks!

PR #478, opened 2023-09-28
add japanese translation

makotot:
Added Japanese translation for https://kentcdodds.com/blog/write-tests


kentcdodds (commented):
Thank you!

PR #479, opened 2023-10-10
feat: use client-hints for theme

kishanhitk:
Use Client Hints polyfill from The Epic Stack to get the user's theme preference. 

Ref: https://github.com/epicweb-dev/epic-stack/blob/main/docs/client-hints.md
Ref: https://x.com/kentcdodds/status/1711082060046373174


kishanhitk:
Should we also add system theme mode, like we have in the epic stack example? 
Or let it be only expliciy light and dark mode like we have currently?

kentcdodds:
I would definitely like to support system

kishanhitk:
To be more specific:
Here are the two approaches:

1. EpicStack example: The theme selector button has three options: dark, light, and system. 
    If you choose either light or dark, then your system theme mode won’t be respected.
    But, if you choose the system mode, it will follow the OS’s theme mode.
    
2. kentcdodds.com's current implementation: The theme selector allows only two options: light and dark.
    But it also respects your system theme mode by default. This means no matter what you selected explicitly from the UI, if you change your OS’s theme later, the website will follow it. To be more clear, the latest toggle will be followed, whether it is from the website's theme selector or a change of OS's theme.
    
    
Which approach do you prefer? I suggest the 2nd because it will fit perfectly with the current theme selector animation (the sunrise and moonrise animation).

kentcdodds:
Yeah, but the 2nd functionality is also unclear. I definitely want the EpicStack's implementation. If you can come up with a nice improvement to the animation that's great. Or you can do a dropdown like the Epic Stack's profile button does.

kishanhitk:
This is what the new animation looks like:

https://github.com/kentcdodds/kentcdodds.com/assets/41117038/ae1f584b-ae3c-4754-96ea-3fabaf02871f

app/entry.server.tsx
```diff
@@ -132,7 +134,7 @@ function serveBrowsers(...args: DocRequestArgs) {
     remixContext,
     loadContext,
   ] = args
-  const nonce = loadContext.cspNonce ? String(loadContext.cspNonce) : undefined
+  const nonce = String(loadContext.cspNonce)
```
kentcdodds: I'm not sure I understand why this file was changed.

app/utils/client-hints.tsx
```diff
@@ -0,0 +1,139 @@
...
+  .map(hint => {
+    const cookieName = JSON.stringify(hint.cookieName)
+    return `{ name: ${cookieName}, actual: String(${hint.getValueCode}), cookie: cookies[${cookieName}] }`
+  })
+  .join(',\n')}
+];
+for (const hint of hints) {
+	if (decodeURIComponent(hint.cookie) !== hint.actual) {
+		cookieChanged = true;
+		document.cookie = encodeURIComponent(hint.name) + '=' + encodeURIComponent(hint.actual) + ';path=/';
```
kentcdodds: I've made a few to this that I'd like applied here: https://github.com/epicweb-dev/epic-stack/commit/3dca3b0dffc1dc84a77e0cdbb306f9397e5a0340

app/utils/theme.tsx
```diff
@@ -0,0 +1,85 @@
...
+  if (serverRenderWithUnknownTheme) {
+    // stick them both in and our little script will update the DOM to match
+    // what we'll render in the client during hydration.
+    return (
+      <>
+        {React.createElement('dark-mode', null, dark)}
+        {React.createElement('light-mode', null, light)}
+      </>
+    )
+  } else {
```
kentcdodds: We now know the theme on the server, so this shouldn't be necessary.

kentcdodds (changes requested):
This is looking great! Just a few things.

Thanks a lot!

app/entry.server.tsx
kishanhitk: I changed it because Eslint was giving some warning here. 
Changed it back to original now.

kishanhitk:
Thanks @kentcdodds,
I have resolved all the changes you asked for.

P.S.- Sorry it took so long; I was on vacation.

app/components/navbar.tsx
```diff
@@ -65,45 +66,76 @@ function NavLink({
...
-    >
-      {/* note that the duration is longer then the one on body, controlling the bg-color */}
-      <div className="relative h-8 w-8">
-        <span
-          className="absolute inset-0 rotate-90 transform text-black transition duration-1000 motion-reduce:duration-[0s] dark:rotate-0 dark:text-white"
-          style={iconTransformOrigin}
-        >
-          <MoonIcon />
-        </span>
+    <fetcher.Form method="POST" action="action/set-theme">
```
kentcdodds:
```suggestion
    <fetcher.Form method="POST" action="/action/set-theme">
```

Otherwise this will be relative to wherever the component is rendered.

kentcdodds (approved):
Awesome! Thank you! This is a huge help :)

kentcdodds:
Looks like there are typescript issues. I don't think they're related to your changes. The unified ecosystem can be a bit of a mess sometimes. They just had a v3 release so I'm guessing these are related to that :(

kentcdodds:
Thank you so much for making this contribution. The fact my website didn't have a system mode really bugged me and I never liked my theming hacks either. I'm definitely happier with this now :) Thanks!

app/utils/theme-provider.tsx
```diff
@@ -1,236 +0,0 @@
...
-    }
-  } else {
-    console.warn(
-      "Heya, could you let Kent know you're seeing this message? Thanks!",
-    );
-  }
-})();
-`
-
-function handleDarkAndLightModeEls() {
```
kentcdodds, line 124: I'm so glad this is gone

PR #482, opened 2023-11-03
feat: add fetcher key for theme fetcher

### LINKED ISSUE #481: Optimistic Theme not working after migrating to client hints
I made a small mistake in my PR #479, because of which the optimistic theme mode is not working.
The URL used to determine the theme change form action is incorrect. It should be `/action/set-theme` and not `/`. 
https://github.com/kentcdodds/kentcdodds.com/blob/e488c740c83da7979adb1fd2366a7933df4a213d/app/utils/theme.tsx#L38

This is a very good example that shows how fetcher identifiers can be useful.

I will create a PR that fixes this, and utilise the new fetcher identifier API.

kishanhitk:
Fixes: #481


kentcdodds (approved):
Thank you! Love this new feature 👏👏

PR #483, opened 2023-11-06
fix rss link

strackfeldt:
Hey, the rss link on https://kentcdodds.com/blog/list was broken, this should fix it


kentcdodds (approved):
Nice! Thanks!

PR #486, opened 2023-11-17
feat(mocks): updates msw to version 2 using intuita #2

r4zendev:
This pull request is the result of running [upgrade-recipe](https://github.com/intuita-inc/codemod-registry/tree/main/msw/2) sequence of codemods. Manual corrections were made to make sure no TypeScript errors remained present.

MSW v2 now uses native browser API as a standard, adhering to the modern standards. Verbosity is lowered, as one might use the HttpResponse helper function alongside with simple key-value configuration object which improves the readability and simplifies MSW's API usage.

Metrics:
**Effort**: Codemod sequence was built in a week, and had been run on 5+ codebases to ensure it covers the most used cases.
**Automation coverage**: covered all the necessary changes. Manual changes had to be done to ensure TypeScript is happy.
**Estimated time save**: 8 hours total. Each mod run performed changes on a set of handlers, covering various use-cases. In order to make this happen manually, one would have to go through the entire documentation and perform manual changes. Documentation also does not cover TypeScript changes.
**Number of files changed**: 9 (11 with package.json)
**FP (unwanted changes)**: no known FPs as of now.
**FN (missed changes)**:
- Current codemod does not cover factory methods, meaning if your msw calls have a wrapper around it, we do not support this use-case at the moment.
- If you were using req.body in your interceptors, this codemod will blindly assume you want `await request.json()` instead of any other type. You will have to correct that manually if you want another body type.

### DIFF
--- a/mocks/convert-kit.ts
+++ b/mocks/convert-kit.ts
@@ -14,7 +14,7 @@ type RequestBody = {
 const convertKitHandlers: Array<HttpHandler> = [
   http.get<any, DefaultRequestMultipartBody>(
     'https://api.convertkit.com/v3/subscribers',
-    (req, res, ctx) => {
+    () => {
       return HttpResponse.json({
         total_subscribers: 0,
         page: 1,
@@ -25,7 +25,7 @@ const convertKitHandlers: Array<HttpHandler> = [
   ),
   http.get<any, DefaultRequestMultipartBody>(
     'https://api.convertkit.com/v3/subscribers/:subscriberId/tags',
-    (req, res, ctx) => {
+    () => {
       return HttpResponse.json({
         tags: [
           {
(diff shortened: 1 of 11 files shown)


kentcdodds (approved):
Thank you for this!

kentcdodds:
Thanks! Could you look into the test failure? Does it run successfully for you locally?

r4zendev:
> Thanks! Could you look into the test failure? Does it run successfully for you locally?

I'm on it. My bad, I have not tested running the playwright tests

kentcdodds:
Thanks!

r4zendev:
@kentcdodds May you please run the workflow again? One of the mods had a bug I had to correct, plus I've booted up the application and manually ran the tests that were failing. One of them wanted `text` body type, after changing that, the test suite passed, however one of them is still failing. I might've set my local environment inappropriately, because I feel like the last failing test should pass. I'd like to confirm that.

r4zendev:
Alright, I've followed the code through, it was all about body types haha
In msw v2 `req.body` is a stream and we have to explicitly cast it to a type we want using Request object methods. So, I went through all the usages of mocked requests that use request body and made sure that the request is being casted to a correct type.
This is something our codemods can't really predict, since it's quite hard to determine which body type the user wants in a particular handler! I will update that in README file of the codemod for future users.
Please run the tests again! Thanks!

kentcdodds:
Thank you!

PR #487, opened 2023-11-29
update example code to reuse prepared statements

ryankshaw:
great article Kent! 

feel free to close this without doing anything, as it's just a tip.

While reading, I noticed in your example code that you create a new sqlite prepared statement every time you `get`/`set`/`delete` from the cache. I don't know if you did that for brevity or if your actual code looks like that, but it would be faster to reuse that prepared statement every time (that's the beauty of prepared statements, [they let you skip the sql string parsing and stuff on subsequent calls](https://www.sqlite.org/c3ref/stmt.html#:~:text=Create%20the%20prepared,or%20more%20times) ) 

So this is just a tip for you to change your app code if that is indeed how you're using it, and maybe change your example code so if people copy/paste it, their app also benefits from just preparing it once.

### DIFF
--- a/content/blog/i-migrated-from-a-postgres-cluster-to-distributed-sqlite-with-litefs.mdx
+++ b/content/blog/i-migrated-from-a-postgres-cluster-to-distributed-sqlite-with-litefs.mdx
@@ -395,31 +395,33 @@ well. There were just a few different (better) option names and I needed to
 write a SQLite adapter:
 
 ```ts
+const preparedGet = cacheDb.prepare(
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Oh, that's a great idea! Care to make this change in the code itself as well? 😅 https://github.com/kentcdodds/kentcdodds.com/blob/7315bef79647b7fcfd2c7e4240d094237ba65fbb/app/utils/cache.server.ts#L58-L116

PR #489, opened 2023-12-12
re-use the same prepared statement over and over again

ryankshaw:
sorry, didn't see your request to change the app code too: https://github.com/kentcdodds/kentcdodds.com/pull/487#pullrequestreview-1756281794

this fixes these spots to re-use the same prepared statement each time. should make things a tiny bit faster


kentcdodds (approved):
Love it. Thanks!

PR #491, opened 2023-12-21
Fix the name of the incredible Disneyland ride

bjohn465:
https://disneyland.disney.go.com/attractions/disneyland/star-wars-rise-of-the-resistance/

The ride is really great, by the way. If you have a chance to ride it, do it.


kentcdodds (approved):
Whoops! Thanks!

PR #494, opened 2023-12-30
Fix duplicate "and" in the problem solution section

ronaldtebrake:
Currently it's 

Screenshot 2023-12-30 at 13 21 12

This should turn it into: 

> Educational blog
> My 200 blog posts (and counting) have been read 717,715 times by 521,596 people. There you'll find blogs about JavaScript, TypeScript, React, Testing, your career, and more.

Maintaining the _and more_ link.


kentcdodds (approved):
Thanks! Amazing how long that was there 😅

PR #495, opened 2024-01-05
Use pure css for hero reveal animation

kof:
Currently to start the animation, the site is waiting for js bundle to load and this results in a significant delay for the user until they can see the animation on a bad network.

I am switching to using pure CSS animation that is preloaded and starts as soon as the html element is rendered.

- uses pure css
- keeps reduced motion preference 

Lets compare before/after pagespeed metrics!


kentcdodds (approved):
HECK YES. Can we do the same for the text as well?

kof:
True, the entire text/cta section has the same problem

PR #496, opened 2024-01-06
Hero reveal animations sequence rewritten to pure CSS with no runtime JS involved

kof:
- Supports reduced motion
- Works with JS disabled
- Doesn't wait for bundle to load

https://github.com/kentcdodds/kentcdodds.com/assets/52824/ff949701-9f3a-49a4-ae7c-a4ee214ed743


kentcdodds (approved):
Let's freaking goooo! Very cool. Thanks for this improvement!

PR #497, opened 2024-01-06
Set loading of the hero image to eager

kof:
On low bandwidth we need to prioritize hero images for faster LCP


kentcdodds:
This conflicted with a few things and it was missing a few spots so I did it myself. Thanks for all the help!

PR #510, opened 2024-03-14
Update use-ternaries-rather-than-and-and-in-jsx.mdx

Helga-UA:
Hello!
I'm Olha from Ukraine 👋
Thank you for the great content!


kentcdodds (approved):
Thanks!

PR #512, opened 2024-03-31
fix: update talk links

alcpereira:
## Problem
While reading "2010s decade in review", I noticed that the link to the talks were not working.

After a quick test, it seems that without the `#` it works as expected.

## Solution
- Replace the `/talks/#my-slug` by `talks/my-slug` links in the MDX files

✅  Tested locally


kentcdodds (approved):
Thanks!

PR #513, opened 2024-04-14
Add translation for "Use react-error-boundary to handle errors in React" in Korean

hatchling13:
Hello, I'm Jung Wook Park from South Korea!
I've added a Korean translation for your post "Use react-error-boundary to handle errors in React".
Thank you for the great article!


kentcdodds (approved):
Thanks!

PR #514, opened 2024-04-24
Update index.mdx, changing 'object' to 'value'

pjvanhall:
the 'reducer' c => c + 1 returns a primitive value of type 'number', not an object


kentcdodds (approved):
Thanks!

PR #521, opened 2024-06-22
Add Chinese (Simplified) translation for get-a-catch-block-error-message-with-typescript.mdx

boris1993:


kentcdodds (approved):
Thank you!

PR #526, opened 2024-11-11
fix link error in a blog from a recent newsletter

thethmuu:
There is a wrong (maybe old) link referenced in a blog post that I found through your recent newsletter and I find and replace with working link.


thethmuu:
BTW, Kent, this is my 2nd OS contribution :D
@kentcdodds

kentcdodds (approved):
Thank you!

PR #534, opened 2025-02-10
Update useeffect-vs-uselayouteffect.mdx

amiryxe:
Add translation for "useEffect vs useLayoutEffect" in Persian


kentcdodds (approved):
Thank you very much!

PR #543, opened 2025-07-04
Fix mobile navigation menu responsiveness

kentcdodds:
Fix mobile navigation menu getting stuck by robustly managing body overflow and closing popover on route change.

The previous implementation had a race condition where the `body:has(#mobile-menu:popover-open) { overflow: hidden; }` CSS rule would sometimes persist `overflow: hidden` on the body even after the popover visually closed, preventing user interaction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the behavior of the mobile menu to prevent background scrolling when the menu is open.
  * Ensured the mobile menu automatically closes when navigating to a new page.

* **Style**
  * Updated CSS to better manage mobile menu transitions and interactions.

* **Chores**
  * Added comments to clarify how overflow behavior is now managed.

### DIFF
--- a/app/components/navbar.css
+++ b/app/components/navbar.css
@@ -1,25 +1,22 @@
-body:has(#mobile-menu:popover-open) {
-	overflow: hidden;
-
-	[popovertarget='mobile-menu'] {
-		rect:nth-of-type(1) {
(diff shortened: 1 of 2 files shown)


kentcdodds:
Yolo

PR #744, opened 2026-03-22
fix: `semantic-release` shoutout link

yamanidev:
Broken link fix in the [How I am so productive](https://kentcdodds.com/blog/how-i-am-so-productive) article, regarding the shoutout to the `semantic-release` package team.


kentcdodds:
Thanks!

